PR #24, opened 2026-03-31
feat: allow adjusting meal/snack time via long press

igorbasko01:


igorbasko01:
@copilot+claude-opus-4.6 I think that it is possible to consolidate the changes into a single function/widget and reuse it over the different pages and widgets.

PR #34, opened 2026-05-19
feat: remove manual overate toggle and derive it dynamically from weight

igorbasko01:
Completely removes the manual overate toggle from UI and database. Instead, calculates the overeating state dynamically based on the day-to-day weight difference (N+1 > N).

### DIFF
--- a/lib/features/days/data/day.dart
+++ b/lib/features/days/data/day.dart
@@ -11,16 +11,10 @@ class FoodDay {
   final List<Food> meals;
   @HiveField(2)
   final List<Food> snacks;
-  @HiveField(3)
-  bool? _overate;
-
-  bool get overate => _overate ?? false;
-  set overate(bool value) => _overate = value;
 
   FoodDay({
     required this.date,
     required this.meals,
     required this.snacks,
-    bool? overate,
-  }) : _overate = overate ?? false;
+  });
 }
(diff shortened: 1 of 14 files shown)


igorbasko01 (commented):
I've reviewed the changes and here are some observations and suggestions for improvement:

1. **Bug in Streak Calculation Loop:**
   In `day_manager.dart`, `getOvereatingStats` iterates over `pastDays` (which are `FoodDay` objects). If a user forgets to open the app on a given day (e.g. Tuesday), a `FoodDay` might not be created for Tuesday. This means Tuesday won't be in `pastDays`. If the user logged their weight on Tuesday and Wednesday (which indicates an overeating event), the loop will skip Tuesday and check Monday next, allowing the clean streak to incorrectly continue. 
   **Fix:** Since overeating is now derived entirely from `WeightManager`, you should iterate backwards through calendar days (Yesterday, Day Before Yesterday, etc.) starting from `today - 1`, instead of iterating over `pastDays`. You can stop the loop when a weight is missing or when you reach the end of `weightManager.history`.

2. **Daylight Saving Time (DST) Bug:**
   In `isOvereaten`, you use `date.add(const Duration(days: 1))` to find the next day. Dart's `Duration(days: 1)` adds exactly 24 hours. On the day DST ends (when a calendar day has 25 hours), adding 24 hours to midnight will result in 11:00 PM of the *same* calendar day. `getWeightForDate` would then look for the same day's weight rather than the next day's.
   **Fix:** Use `DateTime(date.year, date.month, date.day + 1)` instead. Dart's `DateTime` handles month/year rollovers and DST boundaries safely.

3. **Loss of Historical Manual Data:**
   In `serialization_service.dart`, the `overate` column was removed. This means if a user previously manually marked days as overeaten but didn't consistently log their weight, importing their old CSV or updating the app will permanently lose that historical context. Just something to keep in mind if users rely on their past data!

Overall, the concept looks great, just needs some minor adjustments to the calendar math and streak iteration logic!

PR #69, opened 2026-07-15
Phase 2: BiteAnalytics — counts, averages, max (bite analytics plan)

igorbasko01:
Implements **Phase 2** of `docs/bite_analytics_plan.md` — the pure, read-time `BiteAnalytics` computation over `BiteRepository` (no meals yet; that's Phase 3).

## What was implemented (keyed to the phase checklist)

- **`BiteAnalytics` class** added to `lib/features/bite/data/bite_analytics.dart`, constructed from a `BiteRepository`, mirroring `WeightAnalytics`' pure-projection shape — holds no state, persists nothing.
- **`minBitesForAverage = 40`** constant.
- **`dailyCounts(from, to)`** — delegates to the repository's day-grouped query (Phase 1).
- **`averagePerDay(from, to)`** — mean over only days with ≥ 40 bites (§5.2); zero and lightly-logged days are excluded from both numerator and denominator. Returns `0` when no day qualifies (per the `Future<double>` signature in §3a).
- **`maxDay(from, to)`** — the peak `DailyBiteCount`, `null` on an empty window; ties broken toward the earliest day (chronological order, first max kept).

## Verification

- Tests added in `test/features/bite/data/bite_analytics_test.dart`, exercised against a real in-memory Drift store: sub-40 and zero days excluded from the average, a window with no qualifying day (average `0`), max with ties (earliest wins), and empty data (max `null`), plus a `dailyCounts` ordering/gap case.
- Flutter is not installed in this web session (per `CLAUDE.md`, the toolchain is not installed here), so `flutter analyze` / `flutter test` were **not** run locally — verification is deferred to the `flutter_ci.yml` checks on this PR.

## Plan doc

Phase 2's checklist (including its Verify bullet) is flipped to `- [x]` in `docs/bite_analytics_plan.md`; nothing else in the doc changed.

https://claude.ai/code/session_01S47DJBm9u17Pba9947JLyL

---
_Generated by [Claude Code](https://claude.ai/code/session_01S47DJBm9u17Pba9947JLyL)_

### DIFF
--- a/lib/features/bite/data/bite_analytics.dart
+++ b/lib/features/bite/data/bite_analytics.dart
@@ -1,3 +1,5 @@
+import 'package:food_locker/features/bite/data/bite_repository.dart';
+
 /// Total bites on a single local calendar day.
 ///
 /// A read-time projection of the append-only bite log: [day] is normalised to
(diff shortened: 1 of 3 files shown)


lib/features/bite/data/bite_analytics.dart
```diff
@@ -22,3 +24,52 @@ class DailyBiteCount {
...
+/// persisting nothing. Every method is a view of the stored `at_ms`
+/// timestamps: the class holds no state and can be reconstructed freely.
+class BiteAnalytics {
+  BiteAnalytics(this._repository);
+
+  final BiteRepository _repository;
+
+  /// Days below this bite count don't count toward [averagePerDay]: a day you
+  /// forgot to log — or only logged a few bites on — never dilutes the mean.
+  static const int minBitesForAverage = 40;
```
igorbasko01, line 40: I think that maybe days with less than 40 bites shouldn't be taken into consideration in any analytics calculations, not only mean. wdyt?

igorbasko01: Keeping the threshold scoped to averages only for now (plan §5.2), for two reasons:

- `dailyCounts` feeds the chart, and Phase 5 deliberately *shows* sub-40 days as muted bars below a 40 reference line rather than hiding them — so the raw daily series has to keep every day.
- `maxDay` is the "best day" headline; showing it even in a lightly-logged window is still informative, and the mean already absorbs the partial-logging noise the threshold is meant to guard against.

Happy to revisit extending it to the aggregate stats (max, and the Phase 7 meals-per-day average) if that reads better once the chart is in — but that'd be a plan change rather than part of this phase.

---

PR #85, opened 2026-07-17
feat: body weight as a second bar on the bite chart (Phase 5)

igorbasko01:
Implements **Phase 5** of `docs/bite_analytics_enhancements_plan.md` — overlaying daily body weight as a second grouped bar on the daily-bites chart.

## What changed, keyed to the phase checklist

- **Inject `WeightRepository` into `BiteAnalyticsController`** and load raw daily weights over the existing 30-day window. `load()` now filters `getAllWeights()` to `[from30, to)` and exposes them as `dailyWeights` (one per weighed day, no gap-filling). The page passes `context.read<WeightRepository>()` into the controller.
- **Second `BarChartRodData` per group in `daily_bites_chart.dart`** for weight, on a secondary right-hand kg axis fitted to the weight min/max (±1 kg padding). The weight rod's `toY` is normalised into the bite chart's `[0, maxY]`; `rightTitles` inverts that mapping back to kg. The weight rod uses the theme's `tertiary` colour, a two-item legend ("Bites" / "Weight (kg)") is shown above the chart, and each rod is drawn at ~half `_barWidth` when weight is present. The tooltip reads the day's kg from the source (not the normalised rod height).
- **Weight bar omitted on days with no weigh-in; bite-only days unchanged.** The x-axis now spans the union of bite days and weigh-in days, so a weigh-in on a day with no bites gets a slot (zero-height bite rod + weight rod). When `weights` is empty the chart is byte-for-byte the previous bites-only chart: single rod per day, full width, no right axis, no legend.

## Verification

Flutter isn't available in this web session (per `CLAUDE.md`, the toolchain isn't installed here), so local `flutter analyze` / `flutter test` are **deferred to the `flutter_ci.yml` PR checks**. Tests added to cover the phase's Verify bullet:

- `daily_bites_chart_test.dart`: bites-only day has a single rod and no right axis; a weighed day gets a second tertiary weight rod, a right axis, and the legend; a weighed-but-not-eaten day shows only the weight bar (zero bite rod); the weight axis fits the range so a 1 kg change is a visible height difference and the lighter day still draws above the floor (not flattened from zero).
- `bite_analytics_controller_test.dart`: `load` reads only weigh-ins inside the 30-day window, and is empty when none were recorded. Existing controller/page tests updated to inject the weight repository.

No deviations from the plan.

https://claude.ai/code/session_01KW4f9mVTf2eCNzT5mbPD3Z

---
_Generated by [Claude Code](https://claude.ai/code/session_01KW4f9mVTf2eCNzT5mbPD3Z)_

### DIFF
--- a/lib/features/bite/data/bite_analytics_controller.dart
+++ b/lib/features/bite/data/bite_analytics_controller.dart
@@ -148,10 +148,7 @@ class BiteAnalyticsController extends ChangeNotifier {
     );
     final fromYear = DateTime(now.year - 1, now.month, now.day + 1);
     _dailyCounts = await _analytics.dailyCounts(from30, to);
-    _dailyWeights = _weightRepository.getAllWeights().where((w) {
-      final day = DateTime(w.date.year, w.date.month, w.date.day);
-      return !day.isBefore(from30) && day.isBefore(to);
-    }).toList();
+    _dailyWeights = _weightRepository.getWeightsInRange(from30, to);
     _averageLast30 = await _analytics.averagePerDay(from30, to);
     _maxLast30 = await _analytics.maxDay(from30, to);
     _averageLastYear = await _analytics.averagePerDay(fromYear, to);
--- a/lib/features/bite/data/bite_analytics_controller.dart
+++ b/lib/features/bite/data/bite_analytics_controller.dart
@@ -1,6 +1,8 @@
 import 'package:flutter/foundation.dart';
 import 'package:food_locker/features/bite/data/bite_analytics.dart';
 import 'package:food_locker/features/bite/data/bite_repository.dart';
+import 'package:food_locker/features/weight/data/weight.dart';
+import 'package:food_locker/features/weight/data/weight_repository.dart';
 
 /// Per-screen state for the Bite Analytics page.
 ///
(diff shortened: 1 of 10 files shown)


lib/features/bite/data/bite_analytics_controller.dart
```diff
@@ -135,6 +148,10 @@ class BiteAnalyticsController extends ChangeNotifier {
     );
     final fromYear = DateTime(now.year - 1, now.month, now.day + 1);
     _dailyCounts = await _analytics.dailyCounts(from30, to);
+    _dailyWeights = _weightRepository.getAllWeights().where((w) {
```
igorbasko01: maybe WeightRepository should expose a weightForDay method? It feels like it leaked implementation detail (Box) into the bite_analytics_controller.

igorbasko01: Good call. Added a day-granular `getWeightsInRange(from, to)` to `WeightRepository`, implemented once on the `WeightRepositoryHelper` mixin (so both the Hive-backed and in-memory impls get it). `BiteAnalyticsController.load()` now just calls `_weightRepository.getWeightsInRange(from30, to)` — no more `getAllWeights()` + manual date filter leaking into the controller. Pushed in a68a6f8, with repository tests for the range boundaries.

---

PR #91, opened 2026-07-18
fix: refresh bite tab count and current meal on tab switch and resume

igorbasko01:
## Problem

The Bite tab showed stale numbers until the app was closed and reopened:

- **Current meal number didn't refresh on tab switch** — a sitting that ended while you were on another tab kept showing its old `This meal: N`.
- **The day count didn't roll over** — when a new day arrived, `Today's Bites` stayed on the previous day's count even after switching tabs; only a full app restart reset it.

Root cause: `BiteManager` recomputed `todayCount` and `currentMealBites` only in `initialize()` (once, at startup) and after each `logBite()`. Nothing re-read the store when returning to the Bite tab or when the app resumed, so the day window and the trailing-meal projection were never recomputed against the current time.

## Fix

- **`BiteManager.refresh()`** — new method that re-reads the store: rolls the day count to the current local day, drops `currentMealBites` to 0 once the sitting has ended, and re-seeds the pacing reference from the last stored bite (restarting or freezing the countdown as appropriate).
- **`BitePage`** — calls `refresh()` when it becomes the visible tab (`didUpdateWidget`, inactive→active) and when the app resumes onto the Bite tab (`didChangeAppLifecycleState`).

## Tests

Added a `refresh` group to `bite_manager_current_meal_test.dart` covering:

- meal ends past the gap threshold → `currentMealBites` becomes 0;
- day rollover → `todayCount` becomes 0;
- `refresh()` notifies listeners so the tab rebuilds.

Flutter isn't installed in the web session used to author this, so `flutter analyze`/`flutter test` run here on CI rather than locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_013zL9GgaDu7PE8LbbYcZDr9)_

### DIFF
--- a/lib/features/bite/data/bite_manager.dart
+++ b/lib/features/bite/data/bite_manager.dart
@@ -112,6 +112,22 @@ class BiteManager extends ChangeNotifier {
     }
   }
 
+  /// Re-reads the store so the surfaced numbers reflect time that passed while
+  /// the screen was not the active tab or the app was backgrounded: the day
+  /// count rolls to the new local day, [currentMealBites] drops to 0 once the
+  /// sitting has ended, and the pacing reference is re-seeded from the last
+  /// stored bite. Cheap enough to run every time the Bite tab becomes visible.
+  Future<void> refresh() async {
+    final now = clock.now();
+    _pacingConfig ??= await _repository.pacingConfigAt(now);
+    await _refreshTodayCount();
+
+    final last = await _repository.lastBite();
+    _lastBiteAt =
+        last == null ? null : DateTime.fromMillisecondsSinceEpoch(last.atMs);
+    _startCountdown();
+  }
+
   /// Records one bite at the current instant — one tap = one bite, persisted
   /// immediately and never blocked — then refreshes today's count and
   /// restarts the pacing countdown from the fresh reference.
--- a/lib/features/bite/data/bite_manager.dart
+++ b/lib/features/bite/data/bite_manager.dart
@@ -125,6 +125,10 @@ class BiteManager extends ChangeNotifier {
     final last = await _repository.lastBite();
     _lastBiteAt =
         last == null ? null : DateTime.fromMillisecondsSinceEpoch(last.atMs);
+    // Recompute the zone and ticker against the current clock: the OS freezes
+    // timers while backgrounded, so a resumed countdown would otherwise show a
+    // stale zone, and a bite that went stale (day rollover, ended sitting)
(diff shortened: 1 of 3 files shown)


lib/features/bite/data/bite_manager.dart
```diff
@@ -112,6 +112,22 @@ class BiteManager extends ChangeNotifier {
...
+  /// stored bite. Cheap enough to run every time the Bite tab becomes visible.
+  Future<void> refresh() async {
+    final now = clock.now();
+    _pacingConfig ??= await _repository.pacingConfigAt(now);
+    await _refreshTodayCount();
+
+    final last = await _repository.lastBite();
+    _lastBiteAt =
+        last == null ? null : DateTime.fromMillisecondsSinceEpoch(last.atMs);
+    _startCountdown();
```
igorbasko01, line 132: Why startCountdown?

igorbasko01: `refresh()` re-seeds `_lastBiteAt` from the store, and `_startCountdown()` is what turns that reference into the visible pacing state — it recomputes `_pacingZone` against the current clock and (re)arms or stops the ticker accordingly. It matters most on the app-resume path: the OS freezes/throttles Dart timers while backgrounded, so without recomputing here a resumed countdown would display a stale zone. It's also what handles the "bite went stale" cases this PR targets — after a day rollover or an ended sitting, the re-seeded reference is now past `b2`, so `_startCountdown()` stops the ticker and lands on the static clear state instead of leaving a stale ticker running. On a plain tab switch (where the ticker was never stopped) it's effectively a harmless restart from the same reference.

I've pushed c862db1 adding an inline comment at the call site to make this clear in the code.

---

igorbasko01: Is it idempotent to run the start timer method?

igorbasko01: Yes. The first thing `_startCountdown()` does is `_stopTicker()`, which cancels the current ticker and nulls it out before it ever creates a new `Timer.periodic`. So no matter how many times it runs, you always end with exactly zero or one ticker — timers can't stack or leak.

It's also idempotent in outcome: the resulting zone and ticker are derived purely from `_lastBiteAt` and `clock.now()`, so running it N times at the same instant leaves the same steady state as running it once. The only repeated side effect is an extra `notifyListeners()` per call (a harmless rebuild); the haptic (`_onReachedClear`) lives in `_tick`, not `_startCountdown`, so it isn't re-fired.

---

PR #96, opened 2026-08-09
feat: show a progress toast while importing or exporting a backup

igorbasko01:
Backup and restore gave no feedback between the tap and the final toast, so a long operation looked like nothing was happening. Both directions now report themselves while they run, and neither claims success for work that didn't happen.

## Changes

**Import**
- `importData` takes an `onRestoreStart` callback, fired once a file is chosen and the restore is about to begin — after the picker closes, so the message never sits behind the file browser.
- It returns whether a restore actually ran. Dismissing the file picker previously still showed "Data imported successfully"; it now shows nothing.

**Export** (added in response to review)
- `exportData` takes an `onShareReady` callback, fired once the zip is written and the share sheet is about to open, so the progress toast is cleared instead of sitting behind the sheet.
- It returns whether there was anything to export. Both stores empty meant an early return with no file, but the page still said "Data exported successfully"; it now says "No data to export".

**Page**
- An indefinite "Importing data..." / "Exporting data..." snackbar with a spinner runs for the duration of the work, swapped for the success or failure toast when it ends.
- A shared busy flag disables both tiles during either operation, so a double-tap can't start a second destructive clear-then-restore.
- The progress snackbar reads its theme from its own build context via a `Builder`, so showing it doesn't depend on the page still being mounted — the `mounted` check narrows to the one `setState` that requires a live element.

## Tests

New `test/ui/pages/settings_page_test.dart` drives the page against a fake `SerializationService` that holds the work open: progress toast then success for both directions, no toast on a dismissed picker, "No data to export" on an empty export, and the failure toast replacing the progress toast.

`flutter analyze` / `flutter test` were not run locally — Flutter isn't installed in the session environment, so CI is the check here.

https://claude.ai/code/session_018b9tJLoUJTiAi1gDjxMMkm

### DIFF
--- a/lib/ui/pages/settings_page.dart
+++ b/lib/ui/pages/settings_page.dart
@@ -10,6 +10,8 @@ class SettingsPage extends StatefulWidget {
 }
 
 class _SettingsPageState extends State<SettingsPage> {
+  bool _importing = false;
+
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
@@ -32,6 +34,7 @@ class _SettingsPageState extends State<SettingsPage> {
             child: Column(
               children: [
                 ListTile(
+                  enabled: !_importing,
                   leading: Icon(Icons.download, color: theme.colorScheme.primary),
                   title: const Text('Export Data', style: TextStyle(fontWeight: FontWeight.bold)),
                   subtitle: const Text('Save your weight history data into a zip file.'),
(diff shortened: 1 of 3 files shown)


lib/ui/pages/settings_page.dart
```diff
@@ -32,6 +34,7 @@ class _SettingsPageState extends State<SettingsPage> {
             child: Column(
               children: [
                 ListTile(
+                  enabled: !_importing,
```
igorbasko01: I assume that exporting should have also a similar approach as import.

lib/ui/pages/settings_page.dart
```diff
@@ -83,6 +71,57 @@ class _SettingsPageState extends State<SettingsPage> {
...
+  Future<void> _importData() async {
+    final messenger = ScaffoldMessenger.of(context);
+    final service = context.read<SerializationService>();
+    var progressShown = false;
+
+    try {
+      final imported = await service.importData(
+        context,
+        onRestoreStart: () {
+          if (!mounted) return;
```
igorbasko01: is it a good practice using `mounted`?

igorbasko01: Yes for `setState`, no for much else — and the original placement was doing more than it needed to, so I've narrowed it in d98b1ff.

`State.mounted` guarding `setState` after an `await` is the documented Flutter idiom: the async work can outlive the widget, and `setState` on a disposed `State` throws. There's no framework-provided alternative for that — the only way to avoid the check is to move the state out of the widget (e.g. into a `ChangeNotifier`), which isn't worth it for a single "busy" flag.

What *was* smelly is that the check wrapped the whole callback, including `showSnackBar`. That was only necessary because the snackbar builder read `Theme.of(context)` from the page's context. So the progress toast now reads the theme from its own build context via a `Builder`:

```dart
onRestoreStart: () {
  progressShown = true;
  messenger.showSnackBar(_progressSnackBar('Importing data...'));
  if (mounted) setState(() => _busy = true);
},
```

Now `mounted` covers exactly the one call that requires a live element, and the toast still appears if the page went away while the picker was open. Same reasoning behind capturing `ScaffoldMessenger.of(context)` before the `await` — the messenger lives above this page in the tree and outlives it, which is also what keeps `use_build_context_synchronously` quiet without sprinkling `context.mounted` around.

---

lib/ui/pages/settings_page.dart
igorbasko01: Agreed — done in d98b1ff.

Export had the same two problems as import: no feedback while the zip is built, and a success toast that fires even when nothing happened (`exportData` returns early when both stores are empty, but the page still said "Data exported successfully").

The shape mirrors import, with the callback marking the other end of the slow part — import's fires when the picker closes and the restore *starts*; export's fires when the zip is written and the share sheet is about to open, so the progress toast is cleared instead of sitting behind the sheet:

- `exportData` now returns `bool` and takes `onShareReady`.
- The page shows "Exporting data..." on tap and swaps it for "Data exported successfully" / "No data to export" / "Export failed: …".
- The busy flag is now shared (`_busy`), so both tiles are disabled during either operation.

Two export widget tests added alongside the import ones.

One thing I did *not* mirror: import stays silent when you dismiss the picker (you cancelled it, you know why nothing happened), whereas an empty export gets an explicit "No data to export" — you asked for a file and didn't get one, so silence would look broken. Happy to make them consistent either way if you'd rather.

---

PR #97, opened 2026-08-17
feat: pick the weight history range from five presets

igorbasko01:
The Weight page listed every entry ever logged. It now shows a bounded window, with the range picked from a menu next to the History heading: last 7 / 30 / 90 days, last 6 months, last year. It defaults to 7 days.

## Changes

- **`DateRange`** (`lib/core/date_range.dart`) — a span of `days` calendar days ending at `to`, or ending today and open above when `to` is null. `lastDays(n)` builds the relative form, which resolves on read so a range held across midnight still means the same thing; `between(from, to)` builds the anchored form a date picker would produce, deriving the span in UTC so a DST boundary can't shave a day off it. `from` is derived, and equality covers both fields.
- **`WeightRepository.getWeightsSince(since)`** — the window query, implemented once in `WeightRepositoryHelper` next to `getWeightsInRange`, day-granular, inclusive, newest first.
- **`WeightManager`** — holds the selected `DateRange` and loads only that window into `_weights`, so `history` stays the single accessor and the manager holds just what it displays. `_reloadHistory` loads and notifies together; `selectHistoryRange` early-returns when the range is unchanged. `history` re-checks the range on read, so the list is right even if the load happened before midnight.
- **`HistoryRangeSelector`** (`lib/ui/widgets/history_range_selector.dart`) — a `PopupMenuButton` over the five presets, with the wording in a `ui/` extension, matching how `PacingZoneStyle` keeps `PacingZone`'s labels out of the data layer. A menu rather than a segmented control: five labels don't fit a phone's width, and a custom range can later join as one more item.
- **`WeightPage`** — heading row carries the selector; the empty state names the active range ("No weight entries in the last 6 months.").

The chart and the list both read `history`, so they move together on the selected range. The stat tiles stay All Time / 30 / 7 — they're analytics over the whole dataset, not the view window.

## Tests

- `test/core/date_range_test.dart` — `from` for relative and anchored ranges; `between` spans (inclusive, single day, time-of-day stripping, across a leap day, and the exact days each preset label claims); `contains` at the lower boundary, open above for relative ranges and closed for anchored ones; equality on both fields.
- `weight_manager_test.dart` — the default range, `initialize` loading the window from the repository, widening and narrowing the range, and the no-notify no-op.
- `weight_repository_test.dart` — `getWeightsSince` boundary, ordering and empty cases.
- `weight_page_test.dart` — an out-of-range entry is hidden, and picking "Last 30 days" from the menu reveals a 20-day-old entry.

## Notes

- The two long presets are day counts, so their boundaries drift against the calendar: 180 days reaches a few days short of "6 months", and 365 days lands a day or two late across a leap year. The labels are approximate; the spans are exact.
- The selection is in-memory and resets to 7 days on relaunch; there's no prefs store in the app yet.
- `HomePage` reads the same `history`, so it now lists weigh-ins from the last 7 days rather than the 7 most recent entries. Left as-is deliberately, pending the Home page redesign.
- Restoring a backup still doesn't refresh `WeightManager` — pre-existing, filed as #98.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GQ91oVei6xYULsUzvbGzF5)_


lib/features/weight/data/weight_manager.dart
```diff
@@ -12,21 +15,39 @@ class WeightManager extends ChangeNotifier {
...
-    _weights = _weightRepository.getAllWeights();
+    _loadHistory();
     notifyListeners();
   }
 
+  /// How far back [history] reaches.
+  HistoryRange get historyRange => _historyRange;
+
+  /// The weigh-ins inside [historyRange], newest first.
   List<Weight> get history {
```
igorbasko01: It feels like it shouldn't constantly sort. And also why it doesn't just return the _weights. Maybe the repository should sort.

lib/features/weight/data/weight_manager.dart
```diff
@@ -12,21 +15,39 @@ class WeightManager extends ChangeNotifier {
 
   // TODO: Add pagination for weight history loading to handle large datasets efficiently.
   List<Weight> _weights = [];
+  HistoryRange _historyRange = HistoryRange.week;
```
igorbasko01: Could it be that the _historyRange might divert from the actual data in _weights?

lib/features/weight/data/weight_manager.dart
```diff
@@ -12,21 +15,39 @@ class WeightManager extends ChangeNotifier {
...
+
+  /// The weigh-ins inside [historyRange], newest first.
   List<Weight> get history {
     final sortedWeights = List<Weight>.from(_weights);
     sortedWeights.sort((a, b) => b.date.compareTo(a.date));
     return sortedWeights;
   }
 
+  void selectHistoryRange(HistoryRange range) {
+    if (range == _historyRange) return;
```
igorbasko01, line 38: Is it a value comparison or some ref comparison?

lib/features/weight/data/weight_manager.dart
```diff
@@ -12,21 +15,39 @@ class WeightManager extends ChangeNotifier {
...
+    _historyRange = range;
+    _loadHistory();
+    notifyListeners();
+  }
+
+  void _loadHistory() {
+    _weights = _weightRepository.getWeightsSince(
+      _historyRange.startingFrom(DateTime.now()),
```
igorbasko01: Feels a bit convoluted and not straightforward. We basically select a range as an enum but then provide a date which also sounds like it should be in the past but we pass current date.

lib/features/weight/data/weight_manager.dart
```diff
@@ -12,21 +15,39 @@ class WeightManager extends ChangeNotifier {
...
   // TODO: Add pagination for weight history loading to handle large datasets efficiently.
   List<Weight> _weights = [];
+  HistoryRange _historyRange = HistoryRange.week;
 
   WeightManager(this._weightRepository)
     : _analytics = WeightAnalytics(_weightRepository);
 
   Future<void> initialize() async {
-    _weights = _weightRepository.getAllWeights();
+    _loadHistory();
```
igorbasko01: Maybe we can create a loadAndNotify method to not by accident miss a notify when loading? Or is it an overkill?

lib/features/weight/data/weight_manager.dart
igorbasko01: Agreed, and done — the repository sorts now.

`getWeightsSince` returns newest first (documented on the interface), so `_weights` is already in display order and `history` is a single in-range pass, no copy-and-sort per rebuild.

It still can't just return `_weights`: the returned list would be the manager's own mutable state, and callers could sort or clear it behind the manager's back. It's now `toList(growable: false)` from the in-range filter, which also gives that protection for free. The filter itself is doing real work — see the midnight case in the other thread.

---

lib/features/weight/data/weight_manager.dart
igorbasko01: Good catch — it could, in one way, and I've fixed it.

The field itself can't drift: `selectHistoryRange` and every mutation path go through `_reloadHistory()`, so the range and the load always move together. What *could* drift was the **window edge**, because it was resolved from `DateTime.now()` at load time. An app left open across midnight kept a `_weights` that started a day too early, and the list showed 8 days while the range said 7.

Fixed by re-checking the range on read (`HistoryRange.covers`). This is sufficient rather than a band-aid because `getWeightsSince` has no upper bound: as the clock advances, `_weights` can only ever be a *superset* of the current window, never missing entries from it, so trimming on read is exact. Covered by the "a day loaded yesterday falls out once the date rolls over" test.

One divergence I have *not* fixed, since it predates this PR and is a different problem: restoring a backup from Settings rewrites the repository without telling `WeightManager`, so `_weights` is stale (deleted entries still listed, restored ones missing) until the next add/edit/delete or an app restart. Want me to open a follow-up issue for that, or fix it here by having the restore flow re-initialize the manager?

---

lib/features/weight/data/weight_manager.dart
igorbasko01: Value, on both readings of the line:

- `range == _historyRange` in `selectHistoryRange` — enum values are canonical instances, so identity and value equality are the same thing here. Two references to `HistoryRange.week` are always the same object; there's no way to construct a second, equal-but-distinct `week`.
- The date comparisons the line feeds into (`isBefore`, and `DateTime ==`) are value comparisons on the instant, not reference ones, so the freshly constructed `DateTime` that `oldestDay` returns compares correctly against stored dates. Worth knowing that `DateTime ==` also requires `isUtc` to match — every date in this feature is local, and `oldestDay` builds a local one, so that holds.

---

lib/features/weight/data/weight_manager.dart
igorbasko01: You're right that it read backwards — `startingFrom(DateTime.now())` sounds like "start the range at now", when what's being passed is the clock and what comes back is a day in the past. Renamed:

```dart
_historyRange.oldestDay(asOf: DateTime.now())   // was startingFrom(now)
_historyRange.covers(weight.date, asOf: now)    // was includes(date, now: now)
```

`asOf:` being named forces the reading "the oldest day this range covers, as of now", so the direction is explicit at the call site.

On why the date is passed in at all rather than the enum reading the clock itself: it's the seam that makes the range testable. `oldestDay` and `covers` are pure functions of (range, instant), so `history_range_test.dart` can pin a date and assert exact days — including the midnight rollover in the other thread, which is otherwise untestable without a fake clock. The alternative, an `oldestDay` getter calling `DateTime.now()` internally, is shorter at the two call sites and costs us that.

---

lib/features/weight/data/weight_manager.dart
igorbasko01: Not overkill — done. `_reloadHistory()` now loads and notifies:

```dart
void _reloadHistory() {
  _weights = _weightRepository.getWeightsSince(
    _historyRange.oldestDay(asOf: DateTime.now()),
  );
  notifyListeners();
}
```

All five call sites (`initialize`, `selectHistoryRange`, `addWeight`, `updateWeight`, `deleteWeight`) are now a single line each, and the pairing can't be broken by a future mutation path forgetting the second half. It also removed the four `// Refresh from repository to ensure consistency` comments, since the method name now says it.

The one thing it gives up: a caller can no longer reload without notifying — e.g. batching several writes and notifying once at the end. Nothing needs that today, and if something does, the honest shape is a separate `_load()` that `_reloadHistory()` wraps rather than un-pairing these.

---

lib/features/weight/data/weight_manager.dart
igorbasko01: Filed as #98 — out of scope for this PR, nothing further to change here.

---

lib/features/weight/data/weight_manager.dart
igorbasko01: Can't we have a way to set the asOf with a null which will mean a now?

igorbasko01: Yes — done in `93aab30`. `asOf` is now optional on both methods, defaulting to the clock:

```dart
DateTime oldestDay({DateTime? asOf}) {
  final on = asOf ?? DateTime.now();
  ...
}
```

The reload is a single line again: `getWeightsSince(_historyRange.oldestDay())`.

One call site still passes it explicitly — `history`:

```dart
final now = DateTime.now();
return _weights.where((w) => _historyRange.covers(w.date, asOf: now)).toList(growable: false);
```

It's judging a batch, so hoisting the instant means every weigh-in is measured against the same moment rather than each one re-reading the clock mid-filter. Documented on `covers` so the next caller knows when passing it is worth it. Tests keep pinning `asOf` for the exact-day assertions, and there are two new ones covering the null default.

---

lib/core/date_range.dart
```diff
@@ -0,0 +1,36 @@
...
+    final on = asOf ?? DateTime.now();
+    return DateTime(on.year, on.month, on.day - (days - 1));
+  }
+
+  /// Whether [date]'s calendar day falls in the range as of [asOf] — as of now
+  /// when [asOf] is null.
+  ///
+  /// Pass [asOf] when judging a batch of dates, so they are all measured
+  /// against one instant.
+  bool covers(DateTime date, {DateTime? asOf}) {
```
igorbasko01: covers? shouldn't it be just in? it seems that we can pass now into asOf. the wording is a bit confusing in my opinion, wdyt?

lib/features/weight/data/weight_repository.dart
```diff
@@ -82,6 +86,17 @@ mixin WeightRepositoryHelper implements WeightRepository {
     }).toList();
   }
 
+  @override
+  List<Weight> getWeightsSince(DateTime since) {
+    final sinceDay = DateTime(since.year, since.month, since.day);
+    final weights = getAllWeights().where((w) {
+      final day = DateTime(w.date.year, w.date.month, w.date.day);
+      return !day.isBefore(sinceDay);
```
igorbasko01, line 94: isn't there isAfter and equals?

lib/core/date_range.dart
igorbasko01: Agreed on the verb — renamed to `contains` in `1ed6b26`. `in` isn't available (Dart reserves it for `for`-`in`), and `contains` is how the rest of the language asks this question — `Iterable.contains`, `Rect.contains` — so the call site reads plainly:

```dart
_historyRange.contains(weight.date, asOf: now)
```

On `asOf` taking "now": that is the normal case, and it's optional precisely so callers who mean now can omit it — `range.contains(date)` reads the clock itself. The one place that still passes it is `history`, which is judging a batch and wants every weigh-in measured against a single instant instead of each call re-reading the clock. So the parameter earns its keep only in that batch case; everywhere else it stays invisible.

If `asOf` is itself the confusing part, `now:` is the other candidate — it'd read `contains(date, now: now)`, which I found worse at the one call site that passes it. Happy to switch if you prefer it.

---

lib/features/weight/data/weight_repository.dart
igorbasko01: There is — `isAfter` plus `isAtSameMomentAs` (there's no `equals`; `==` works too, though it also compares `isUtc`). So the explicit form would be:

```dart
return day.isAfter(sinceDay) || day.isAtSameMomentAs(sinceDay);
```

That's exactly equivalent to `!day.isBefore(sinceDay)` for these values — both are local, non-null `DateTime`s normalised to midnight. I picked `!isBefore` for two reasons: it matches `getWeightsInRange` directly above it, which already writes its inclusive lower bound as `!day.isBefore(fromDay)`, and it's one comparison rather than two.

The inconsistency you're seeing is real, though, and it's older than this PR: `getLowestWeight` in the same mixin uses the `isAfter || isAtSameMomentAs` form for the same "since, inclusive" question. So the file currently says it both ways.

Which way do you want it? I'd unify on `!isBefore` — three sites, all inclusive lower bounds — but that means touching `getLowestWeight`, which this PR otherwise doesn't. Say the word and I'll do it here, or I'll leave it and we can tidy it separately.

---

igorbasko01: leave it.

lib/core/date_range.dart
igorbasko01: now I understand what confuses me, history is a range, so when writing contains with two parameters, it becomes not clear what is checked within the range. why do we need this asOf at all? doesn't the range specifically should look at `now` as the upper bound? when is it not the case?

igorbasko01: You've got it right, and I've dropped it — `7deaafd`.

To answer directly: the upper bound is **always** now-and-beyond, and there's no case where it isn't. `asOf` never moved the upper bound — it moved the whole window, since it's the day the count reaches back *from*. So on a two-argument `contains(date, asOf: x)` the second argument silently redefines the range you're testing against, which is exactly the ambiguity you're describing.

It now reads:

```dart
bool contains(DateTime date) {
  final day = DateTime(date.year, date.month, date.day);
  return !day.isBefore(oldestDay());
}
```

and the manager is a plain `_weights.where((w) => _historyRange.contains(w.date))`.

`asOf` stays on `oldestDay`, where "counting back from which day" *is* the question, and where the tests pin it to assert exact boundaries — including the rollover case that used to be tested through `contains`.

What I gave up, for the record, since I'd argued for it earlier: `history` no longer measures every weigh-in against one hoisted instant, so a midnight tick landing mid-filter could judge one entry against the new day. The window is day-granular and the next rebuild corrects it, so that was a much smaller benefit than the confusion it was causing at the call site — your read was better than mine.

The one caller a future custom range would bring — absolute `from`/`to` — wouldn't want `asOf` at all, so this doesn't cost anything there either.

---

igorbasko01: I think that the current implementation builds only on now - some number of days, basically it could have built the date range with two dates and thats it. Then we can have helper builders to build the ranges.

igorbasko01: Agreed on the value type — two dates plus builders is the better `DateRange`, and it makes `contains` clock-free:

```dart
class DateRange {
  const DateRange({required this.from, this.to});   // to == null → open-ended
  factory DateRange.lastDays(int days, {DateTime? asOf}) => ...;

  final DateTime from;
  final DateTime? to;

  bool contains(DateTime date) { /* pure: no clock at all */ }
}
```

Two things to decide before I write it, because they follow from "two dates and that's it":

**1. `to` has to be nullable, not today.** If a preset builds `to: today`, a weigh-in you dated tomorrow disappears from the list. Today's behaviour keeps it visible, so presets build open-ended and only a custom range sets `to`.

**2. The manager can no longer store the built range.** A materialized range is a snapshot: held across midnight it means yesterday's window, which is the bug this PR fixed. So the *rule* has to live somewhere that outlives the read, and it can't be inside the range any more — the manager stores `int _historyDays = 7` and builds `DateRange.lastDays(_historyDays)` in `_reloadHistory` and in `history`. That also gets back the batch consistency I gave up in `7deaafd`: one clock read per pass, one range, `contains` pure.

That forces one call: **what does the selector hand back?** A `DateRange` can't say "I'm the 7-day preset" — you'd have to reverse it out of `to.difference(from)` — so the menu's value type becomes the rule, i.e. `int days`, and the label switch moves onto that. `HistoryRangeSelector` becomes `selected: int` / `ValueChanged<int>`.

So: `int` as the selection type across manager → page → selector, `DateRange` as the transient query value. Fine by me — it's honest about which one is remembered and which one is derived — but it's a different public shape from what's on the branch now, so I'd rather you okay it than discover it in a diff. Say go and I'll push it; say "later" and it rides along with the custom picker, where `DateRange.between` shows up anyway.

For status: CI is green on `7deaafd`.

---

igorbasko01: Can we hold instead of two dates, a single date named `to` and a number of days with direction?

igorbasko01: That's better than the two-date version, and it dissolves the objection I raised in my last comment.

The reason: with a **nullable** anchor, one stored value can be either kind. `to == null` means "ends today, resolved on read" — so a preset stays relative and survives midnight — while a set `to` is an absolute custom range. That's the thing two absolute dates couldn't do, so the manager and the selector keep holding a `DateRange`, and none of the `int days` plumbing from my last comment is needed. What's on the branch now is the same idea with the anchor left implicit; your version just makes it a field.

```dart
class DateRange {
  const DateRange.lastDays(this.days) : to = null;          // relative, follows the clock
  const DateRange.endingAt(this.to, {required this.days});  // absolute

  final DateTime? to;
  final int days;
}
```

Two notes on the semantics:

**Future-dated entries.** A closed upper bound would drop them: date a weigh-in tomorrow and it vanishes from the list — including from the row you'd use to fix it. So a relative range (`to == null`) should stay open above, exactly as it is today, and only an anchored range closes at both ends. That makes `to` carry two meanings — "ends today" and "no ceiling" — which I'd document rather than add a third field for.

**Direction** I'd leave out. Nothing in the app looks forward from an anchor, and adding it makes every consumer and `contains` handle a case that doesn't exist. "The `days` days ending at `to`" is the whole semantic, and the constructor names carry it. Easy to add later if a forward range ever shows up.

On timing: the `to` field earns nothing until the picker exists, so my instinct is to land #97 as-is and add the field in that PR, where `endingAt` gets its first caller and its first test. But it's a small, self-contained addition — say the word and I'll put it in now instead.

---

igorbasko01: I think the `to` and number of days is an internal implementation, we don't have to use the same on the constructors. I think that instead of endingAt we can create a daterange constructor with two dates, and i ternally it will translate it to `to` and number of days. In my opinion endingAt is not clear enough.

igorbasko01: Right — the constructors are the API, `(to, days)` is just how it's kept. `DateRange.between(from, to)` says what the caller means; `endingAt` made them think about the storage. So:

```dart
const DateRange.lastDays(this.days) : to = null;          // relative, open above
factory DateRange.between(DateTime from, DateTime to);    // absolute, closed both ends
DateTime get from => oldestDay();                          // derived, for the picker to show
```

One trap in that translation, since it's the reason to write it once inside the class rather than at each call site: computing the span as `to.difference(from).inDays + 1` is **wrong across a DST boundary**. Local midnights an exact 90 days apart can sit 89 days 23 hours apart in absolute time, and `inDays` truncates, so you'd store 89. The fix is to do the subtraction in UTC:

```dart
final span = DateTime.utc(to.year, to.month, to.day)
    .difference(DateTime.utc(from.year, from.month, from.day))
    .inDays + 1;
```

UTC has no DST, so every calendar day is exactly 24 hours and the count is exact. Same reasoning as the `DateTimeCalendar` note in `weight_analytics.dart`, just in the other direction — that one steps days without arithmetic, this one measures them without local offsets.

That settles the shape. Since none of it has a caller until the picker exists, I'll build it there — `between`, the `from` getter, the UTC span, and their tests — unless you'd rather I add it here.

---

igorbasko01: Add it here because we have the whole context and reasoning here. Don't write comments explaining why or anything similar.

igorbasko01: Added in `e919541`, no explanatory comments — just the two-line doc stating what the class holds.

```dart
const DateRange._(this.to, this.days) : assert(days > 0);
const DateRange.lastDays(int days) : this._(null, days);
factory DateRange.between(DateTime from, DateTime to);

final DateTime? to;
final int days;
DateTime get from => oldestDay();
```

`oldestDay` counts back from `to ?? asOf ?? now`, so an anchored range ignores the clock entirely. `contains` keeps the lower bound as it was and adds the ceiling only when there's an anchor. Equality and `hashCode` now cover both fields, so an anchored 10-day range is not equal to `lastDays(10)` — which is what keeps the menu's selected-item check honest once custom ranges join it.

Nothing else changed: the manager and the selector still hold a `DateRange`, and the presets are still `const`.

Nine new tests — inclusive span, single day, time-of-day stripping, a span across the leap day, an anchored range unmoved by `asOf`, the closed upper bound, and equality on both fields.

---

lib/core/date_range.dart
```diff
@@ -0,0 +1,44 @@
...
+        1;
+    return DateRange._(DateTime(to.year, to.month, to.day), span);
+  }
+
+  final DateTime? to;
+  final int days;
+
+  DateTime get from => oldestDay();
+
+  DateTime oldestDay({DateTime? asOf}) {
```
igorbasko01: Why do we need the asOf here? We have the full date range, so we know the oldest day.

lib/core/date_range.dart
igorbasko01: Gone — `e1db74f`. And `oldestDay` went with it: once the anchor is a field, that method was just a long way of saying `from`.

```dart
DateTime get from {
  final end = to ?? DateTime.now();
  return DateTime(end.year, end.month, end.day - (days - 1));
}
```

Public surface is now `from`, `to`, `days`, `contains` — no clock parameter anywhere. The manager reads `getWeightsSince(_historyRange.from)`.

The reason `asOf` was still there was the tests: it let them pin a date and assert that "180 days" reaches 2023-11-23. That coverage survives without it, because `between` is the same arithmetic in reverse — instead of "count 180 back from May 20", the test now asserts `between(2023-11-23, 2024-05-20).days == 180`. Same fact, both preset drift cases still pinned to explicit dates, and no test-only parameter on the public API. Relative ranges keep their today-based tests.

---

lib/core/date_range.dart
igorbasko01: We don't — it's gone in `e1db74f`. (Apologies: I answered this in the older `contains` thread instead of here.)

You're right that the range knows its own oldest day. `asOf` only ever mattered for a relative range, and once `to` is a field the method can decide for itself, so `oldestDay` collapsed into a getter:

```dart
DateTime get from {
  final end = to ?? DateTime.now();
  return DateTime(end.year, end.month, end.day - (days - 1));
}
```

The public surface is now `from`, `to`, `days`, `contains`, with no clock parameter anywhere, and the manager reads `getWeightsSince(_historyRange.from)`.

The one thing `asOf` was still buying was test precision — pinning a date to assert "180 days back from May 20 is Nov 23". That survives without it, because `between` is the same arithmetic inverted: `between(2023-11-23, 2024-05-20).days == 180`. Both drift cases stay pinned to explicit dates, and no test-only parameter remains on the API.

CI is green on `e1db74f`.

---

lib/features/weight/data/weight_manager.dart
igorbasko01: I assume this one is not relevant anymore, should be resolved?

igorbasko01: Yes — obsolete twice over, and resolved. `startingFrom` became `oldestDay(asOf:)`, then the parameter went away entirely, and the method with it: `DateRange` now exposes a `from` getter and no clock argument anywhere.

---

PR #109, opened 2026-08-25
fix: refresh the managers after a backup restore

### LINKED ISSUE #98: Restoring a backup leaves the Weight tab showing pre-import data
Split out of a review thread on #97 ([discussion](https://github.com/igorbasko01/food-locker/pull/97#discussion_r3796855408)). Predates that PR — the window/range work there only made it easier to notice.

## What happens

Importing a backup from the Settings tab writes straight to the stores and never tells the managers, so `WeightManager._weights` still holds whatever was loaded before the import. Switching to the Weight tab shows the old list: weigh-ins the import deleted are still listed, and restored ones are missing.

`SerializationService.importData` reads the repositories out of the context and calls `restoreFromBackup(weightRepo, biteRepo, bytes)` (`lib/features/settings/data/serialization_service.dart:100-117`), which clears and repopulates them. Nothing in that path touches `WeightManager`, and `initialize()` only runs at startup from `main.dart`.

## Steps to reproduce

1. Log a few weigh-ins.
2. Settings → import a backup zip whose `weight.csv` holds different entries.
3. Go to the Weight tab (without restarting the app).

The list, chart, and stat tiles all reflect the pre-import data. Any add/edit/delete refreshes it, as does restarting the app — which is why this hides easily.

## Notes

- The stat tiles and analytics read the repository at access time, so they are correct as soon as the tab rebuilds; it's the manager's cached `_weights` that goes stale, so the tab can show a stale list next to fresh stats.
- The Bite tab looks less exposed: `BitePage` calls `BiteManager.refresh()` on lifecycle resume (`lib/ui/pages/bite_page.dart:79`), so its count re-reads the store on the next resume. Worth confirming while fixing this rather than assuming.

## Possible fix

Give the restore path a way to tell the managers to reload — e.g. `importData` calling `WeightManager.initialize()` (and `BiteManager.refresh()`) after `restoreFromBackup` returns, or having the Settings page do it once the import reports success. Wiring the managers into `SerializationService` widens its dependencies, so doing it from the page may be the smaller change; either way it wants a test that a restore makes the manager reflect the new store contents.

igorbasko01:
## What changed

`SerializationService.restoreFromBackup` writes straight through the repositories, so `WeightManager._weights` kept serving whatever it had loaded before the import. The Weight tab listed weigh-ins the import had deleted and hid restored ones, next to stat tiles that were already correct (they read the repository at access time).

- **`lib/ui/pages/settings_page.dart`** — `_importData` now reads `WeightManager` and `BiteManager` out of the context (before the first `await`, alongside `messenger`/`service`) and refreshes both once the import reports success, just before the success snackbar.
- **`lib/features/weight/data/weight_manager.dart`** — new `refresh()`, the public name for the existing `_reloadHistory()` reload path.
- **`lib/features/bite/data/bite_manager.dart`** — `refresh()` re-reads the pacing config instead of keeping any version it already held (`_pacingConfig ??=` → re-read, falling back to the held value if the store has none).

## How the issue's asks are met

- **The Weight tab reflects the restored store without a restart** — the restore path itself refreshes the manager, so it no longer depends on a later add/edit/delete or an app restart.
- **"The Bite tab looks less exposed — worth confirming rather than assuming."** Confirmed: `BitePage` refreshes on lifecycle resume *and* on becoming the visible tab (`didUpdateWidget`, `bite_page.dart:56-63`), and `AppShell` keeps every tab in an `IndexedStack`, so reaching the Bite tab after an import always re-reads today's count and the last bite. That masks the staleness rather than removing it, so the restore path now refreshes `BiteManager` too — correctness no longer rests on which tab you happen to arrive from.
- **A test that a restore makes the manager reflect the new store contents** — see below.

- **Refresh from the page, not from `SerializationService`.** The issue offered both; the page is the smaller change and keeps the service's dependencies as they are (it takes repositories, not managers), leaving `restoreFromBackup` unit-testable on its own.
- **A new `WeightManager.refresh()` rather than calling `initialize()` post-startup.** One extra three-line method, but it mirrors the `BiteManager.refresh()` that already exists for exactly this, and stops `initialize()` from acquiring a second meaning that a later reader would reasonably "clean up".
- **Refreshing runs inside the existing `try`.** A failing reload therefore surfaces as the "Import failed" toast even though the stores were replaced. Both reloads are repository reads that shouldn't throw in practice, and the alternative — swallowing the error and reporting success — hides a genuinely broken state.
- **The pacing-config re-read is included.** Strictly it's a second staleness bug in the same restore path (a restore replaces the config history, but `refresh()` kept the thresholds it already had until the next app start). It's one line and squarely the issue's subject — stale in-memory state after a restore — so fixing it here rather than leaving a known hole in the fix. Flagging it explicitly since it's the one part not named in the issue body. The re-read falls back to the held config when the store has none, so it never regresses to the hardcoded fallback thresholds.
- **Both managers are read before the first `await`**, matching how the method already captures `messenger` and `service`, so nothing reads from a `BuildContext` across an async gap.

## Tests

- `test/features/weight/weight_manager_test.dart` — new `WeightManager refresh` group: picks up an entry written straight to the repository, drops one the repository no longer holds, and notifies listeners so the tab rebuilds.
- `test/ui/pages/settings_page_test.dart` — new `leaves no manager holding pre-import data`: drives the real `SettingsPage` through a confirmed import whose fake restore replaces the stores behind both managers' backs, then asserts `WeightManager.history` carries the restored weigh-in and `BiteManager.b2` carries the restored threshold. `pumpPage` now supplies both managers (defaulting to an `InMemoryWeightRepository` and an in-memory fake `BiteRepository`), so the existing import/export tests cover the new code path too.
- `test/features/bite/data/bite_manager_pacing_test.dart` — new `refresh adopts thresholds that were replaced in the store`.

## Verification

Flutter is not installed in this environment and `CLAUDE.md` says not to install it here, so `flutter analyze` and `flutter test` were **not** run locally — the `flutter_ci.yml` checks on this PR are the verification. The diff was kept small and self-reviewed for unused imports and dead code.

Closes #98

---
_Generated by [Claude Code](https://claude.ai/code/session_01EACQpaencGUh158V6XDxKA)_

### DIFF
--- a/lib/features/bite/data/bite_manager.dart
+++ b/lib/features/bite/data/bite_manager.dart
@@ -117,9 +117,12 @@ class BiteManager extends ChangeNotifier {
   /// count rolls to the new local day, [currentMealBites] drops to 0 once the
   /// sitting has ended, and the pacing reference is re-seeded from the last
   /// stored bite. Cheap enough to run every time the Bite tab becomes visible.
+  ///
+  /// The thresholds are re-read on every call: a backup restore can replace the
+  /// config history underneath the manager.
   Future<void> refresh() async {
     final now = clock.now();
-    _pacingConfig ??= await _repository.pacingConfigAt(now);
+    _pacingConfig = await _repository.pacingConfigAt(now) ?? _pacingConfig;
     await _refreshTodayCount();
 
     final last = await _repository.lastBite();
--- a/lib/features/weight/data/weight_manager.dart
+++ b/lib/features/weight/data/weight_manager.dart
@@ -22,6 +22,13 @@ class WeightManager extends ChangeNotifier {
     _reloadHistory();
   }
 
+  /// Re-reads the store, for callers that replaced its contents without going
+  /// through this manager — a backup restore. Ordinary mutations reload
+  /// themselves.
+  Future<void> refresh() async {
+    _reloadHistory();
+  }
+
   DateRange get historyRange => _historyRange;
 
   /// The weigh-ins inside [historyRange], newest first.
--- a/lib/ui/pages/settings_page.dart
+++ b/lib/ui/pages/settings_page.dart
@@ -1,5 +1,7 @@
 import 'package:flutter/material.dart';
+import 'package:food_locker/features/bite/data/bite_manager.dart';
 import 'package:food_locker/features/settings/data/serialization_service.dart';
+import 'package:food_locker/features/weight/data/weight_manager.dart';
 import 'package:provider/provider.dart';
 
 class SettingsPage extends StatefulWidget {
@@ -85,6 +87,8 @@ class _SettingsPageState extends State<SettingsPage> {
   Future<void> _importData() async {
     final messenger = ScaffoldMessenger.of(context);
     final service = context.read<SerializationService>();
+    final weightManager = context.read<WeightManager>();
+    final biteManager = context.read<BiteManager>();
     var progressShown = false;
 
     try {
@@ -101,6 +105,11 @@ class _SettingsPageState extends State<SettingsPage> {
       // removing unconditionally would cut short whatever else is on screen.
       if (progressShown) messenger.removeCurrentSnackBar();
       if (imported) {
+        // A restore writes through the repositories, leaving both managers on
+        // what they loaded before it. Re-read here rather than leaning on a tab
+        // refreshing when it becomes visible.
+        await weightManager.refresh();
+        await biteManager.refresh();
         messenger.showSnackBar(
           const SnackBar(content: Text('Data imported successfully')),
         );
--- a/test/features/bite/data/bite_manager_pacing_test.dart
+++ b/test/features/bite/data/bite_manager_pacing_test.dart
@@ -184,6 +184,27 @@ void main() {
       manager.dispose();
     });
   });
+
+  test('refresh adopts thresholds that were replaced in the store', () {
+    fakeAsync((async) {
+      final repo = _FakeBiteRepository(config: config);
+      final manager = BiteManager(repo, onReachedClear: () async {});
+
+      manager.initialize();
+      async.flushMicrotasks();
+      expect(manager.b2, const Duration(seconds: 30));
+
+      // What a restore does: the config history is replaced underneath the
+      // manager.
+      repo.config = const PacingConfig(id: 2, effectiveMs: 0, b1S: 40, b2S: 90);
+      manager.refresh();
+      async.flushMicrotasks();
+
+      expect(manager.b2, const Duration(seconds: 90));
+
+      manager.dispose();
+    });
+  });
 }
 
 /// An in-memory [BiteRepository] with synchronous-ish futures so the pacing
@@ -192,7 +213,7 @@ class _FakeBiteRepository implements BiteRepository {
   _FakeBiteRepository({this.config, List<DateTime>? bites})
       : _bites = [...?bites];
 
-  final PacingConfig? config;
+  PacingConfig? config;
   final List<DateTime> _bites;
 
   @override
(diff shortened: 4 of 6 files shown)


igorbasko01:
**Verdict: nothing blocks merge.** The fix does what the title says, it's wired the way this repo wires things, and the tests cover the actual behaviour rather than the fact that a file was touched. Two minor notes below, neither of which I'd hold the PR for.

## Invariants checked

- **Generated code / Hive typeIds / Drift schema** — untouched. No model, `@HiveType`, `@HiveField`, or bite table changed, so no stale `*.g.dart` and no `schemaVersion` bump owed. `hive_ce_migration_test.dart` is not modified.
- **Mutation shape** — `WeightManager.refresh()` (`lib/features/weight/data/weight_manager.dart:28`) goes through the one load path, `_reloadHistory()`, which re-reads the repository and notifies. Nothing mutates `_weights` directly. The `refresh()` body is synchronous inside its `Future`, which is correct here: `getWeightsSince` is a synchronous repository read, so the awaited `refresh()` at `settings_page.dart:111` has genuinely finished reloading by the time the success toast goes up.
- **Derived, not stored** — no bite count, delta, or `PacingZone` is persisted. The pacing-config change reads more, it doesn't write.
- **Interfaces over stores** — both managers come out of the existing `MultiProvider` via `context.read`, captured before the first `await`, so no `BuildContext` crosses an async gap. `SerializationService`'s dependencies are unchanged, so `restoreFromBackup` stays unit-testable on repositories alone.
- **Backup compatibility** — no codec, CSV column, or `restoreFromBackup` behaviour changed; the null-vs-empty tolerance and the dedupe are untouched.
- **Versioning** — no manual `pubspec.yaml` / `CHANGELOG.md` edit, and `fix:` is the right prefix for a user-visible bug fix (patch).
- **Coverage** — the manager-level tests and the `SettingsPage` widget test that drives a confirmed import through the real page are the right pair; the widget test would fail against `main`.

The pacing-config re-read (`bite_manager.dart:125`) is a real second bug fixed here, and the `?? _pacingConfig` fallback is the right call — dropping to the hardcoded `_fallbackB1`/`_fallbackB2` when the store happens to hold nothing would be a regression.

## Minor

**1. The restore can start a pacing countdown — and fire the haptic — while the user is on the Settings tab.** `settings_page.dart:112` → `BiteManager.refresh()` → `_startCountdown()` (`bite_manager.dart:135`). If the restored `lastBite()` is within `b2` of now, a `Timer.periodic` starts and `_onReachedClear()` fires when the gap reaches `b2` — in production that default is `HapticFeedback.mediumImpact`.

Failure case: export a backup a few seconds after your last bite, then import that zip from Settings. The phone buzzes on the Settings screen with no bite tapped, and the ticker runs on a tab that isn't visible. Pre-existing `refresh()` semantics, but this PR introduces the call site that can hit them from a non-Bite tab. Rare enough that it's a note, not a blocker; if you want it gone, seeding the reference without restarting the countdown on this path would do it.

**2. A throw in either `refresh()` reports "Import failed" after the stores were already replaced.** Called out in the PR body, and I agree with the reasoning — both are repository reads that shouldn't throw, and reporting success over a broken in-memory state is worse. Noting it only so the tradeoff is on the record in the review too.

## Not findings, for the record

- `WeightManager.refresh()` and `initialize()` now have identical bodies. The PR body explains the choice and the doc comments distinguish them; keeping `initialize()` single-meaning is the better of the two options.
- Another private `_FakeBiteRepository` in `settings_page_test.dart` matches how the other five bite test files already do it, so it isn't new duplication.
- `analyze_and_test` is green on `503a0d5`.

---

igorbasko01 (commented):
Lgtm

Issue #103, opened 2026-08-23
Remove the Home page streak banners and the overeating/streak analytics behind them


igorbasko01:
## Summary

The Home page shows two banners above the weight history — the current-streak banner and the all-time-best banner. Neither earns its place, and the second one is actively misleading. Remove both banners and the whole overeating/streak calculation they are the only consumers of, down to the repository method that exists only to feed it.

```dart
// lib/ui/pages/home_page.dart:50-55
StreakBanner(stats: stats),
LongestStreakBanner(stats: stats),
```

## Why

**The all-time-best banner reports on data that isn't interesting.** `LongestStreakBanner` renders `stats.longestCleanStreakLength` as "All-Time Best: N Days" with the subtitle "Your longest streak of eating mindfully!". Nothing about eating is measured. `WeightAnalytics._isOvereaten(date)` (`weight_analytics.dart:53`) just compares the next day's weigh-in to the current one and calls any increase "overeating". A day's weight is up for plenty of reasons that have nothing to do with what was eaten — hydration, salt, time of day, scale variance — so the number is a proxy for day-to-day scale noise dressed up as a behavioural achievement. It is also a record that can only be beaten by a run of consecutive non-increasing days, which gets rarer the longer the history is, so for most users it becomes a fixed number from months ago that never changes again.

**The current-streak banner doesn't work correctly.** `_calculateCurrentStreak` (`weight_analytics.dart:88`) starts at *yesterday* and asks `_isOvereaten(yesterday)`, which needs **both** yesterday's and **today's** weigh-in. Until today's weight is logged, that returns `null`, the loop breaks on the first iteration, and `currentStreakType` stays `null` — so `StreakBanner` falls through to "Start a new streak! / Log your weight daily to build your momentum." That's the state the banner is in every day before the user weighs in, no matter how long the actual run is. A 30-day streak reads as "no streak" each morning and then reappears after logging.

Two smaller problems in the same code:

- Any missing day anywhere in the run returns `null` and truncates the streak silently, so a single skipped weigh-in resets the displayed streak to zero.
- `currentStreakLength` / `longestCleanStreakLength` (`weight_analytics.dart:29`, `:34`) measure with `end.difference(start).inDays + 1`. That's exactly the DST-unsafe duration math the file's own `DateTimeCalendar` doc comment (`weight_analytics.dart:6`) warns about — a streak spanning a DST transition can come out a day short.

[...]

PR #110, opened 2026-08-26
fix: name the selected history range on Home (#99)

### LINKED ISSUE #99: Home page heading and empty state should name the selected history range
Follow-up to #97, which gave the weight history a selectable range (7 / 30 / 90 days, 6 months, a year). The Weight tab names the active range in its heading and empty state; the Home tab still hardcodes 7 days, so it can now describe itself wrongly.

## What to change

`lib/ui/pages/home_page.dart`:

- **Heading** (`home_page.dart:63`) — `'Latest 7 Days of Weight'` is fixed text. It should follow the selected range, e.g. "Latest 30 Days of Weight" / "Latest 6 Months of Weight". `HistoryRangeLabel.label` in `lib/ui/widgets/history_range_selector.dart` already renders the wording ("Last 30 days", "Last 6 months", "Last year") and could be reused or adapted.
- **Empty state** (`home_page.dart:108-112`) — `'No weight entries yet.\nTap "Log Weight" above to get started!'` fires whenever the window is empty, so it claims there are no entries at all when there may be years of them just outside the range. It should read as "No entries in the last 30 days." and name the active range.

## Worth deciding while doing it

- **Keep the onboarding copy for a genuinely empty store.** "No entries in the last 7 days" is accurate but unhelpful on first run. Two cases: nothing ever logged → keep the "Tap Log Weight above to get started!" prompt; nothing *in the window* → name the range. The manager can distinguish them (the repository is empty vs. `history` is empty).
- **The 7-entry cap.** `home_page.dart:20` takes `history.take(7)`, so with a 30-day range selected the section still shows at most 7 rows under a "Latest 30 Days" heading. Either drop the cap, or word the heading so it describes what's actually listed.
- **Home has no range control.** The range is picked on the Weight tab, so Home's wording would change because of an action on a different screen. Naming the range is still better than misreporting it, but the honest fixes are either giving Home its own control or returning it to "the 7 most recent entries" (its behaviour before #97) — a call for the Home page redesign.

## Context

Before #97, `WeightManager.history` returned every weigh-in, so Home's `take(7)` meant "the 7 most recent entries" regardless of age. `history` is now the selected window, so Home lists only entries inside it — which is what makes both strings above capable of lying.

igorbasko01:
## What changed

The Home tab hardcoded `'Latest 7 Days of Weight'` and an empty state that claimed there were no entries at all, so after #97 gave the history a selectable range both strings could misdescribe what Home was showing.

- **`lib/ui/widgets/history_range_selector.dart`** — one `span` getter on the `HistoryRangeSpan` extension (`7 Days`, `30 Days`, `90 Days`, `6 Months`, `Year`), title-cased and without a baked-in "Last". Call sites frame it: `'Last ${range.span.toLowerCase()}'` for prose, `'Latest ${range.span} of Weight'` for the Home heading.
- **`lib/ui/pages/home_page.dart`** — the heading follows the selected range; the empty state names the active range when the store holds entries outside the window, and keeps the onboarding prompt when the store is genuinely empty; `history.take(7)` is gone.
- **`lib/features/weight/data/weight_manager.dart`** — `hasAnyWeights`, so the page can tell "nothing logged ever" from "nothing in this window" without reaching past the repository interface.
- **`lib/features/weight/data/weight_repository.dart`** (+ both implementations) — `isEmpty` on the interface, answered from the store's own bookkeeping: the Hive box's key count, the in-memory repository's map. Home asks on every rebuild, so it must not deserialize the box.
- **`lib/ui/pages/weight_page.dart`** — its empty state supplies its own "last", following the getter change. No rendered string changes.

## How the issue's asks are met

- **Heading follows the selected range** — `'Latest 30 Days of Weight'`, `'Latest 6 Months of Weight'`, `'Latest Year of Weight'`. With the default 7-day range the string is byte-identical to before.
- **Empty state no longer claims there are no entries at all** — with entries outside the window it reads `No weight entries in the last 30 days.` / `Tap "Log Weight" above to add one.`, phrased to match the Weight tab's equivalent. An empty store still gets `No weight entries yet.` / `Tap "Log Weight" above to get started!`.

## Decisions the issue left open

- **The 7-entry cap → dropped.** The alternative was wording the heading around the cap, but "Latest 30 Days of Weight" above at most 7 rows is the same class of lie the issue is about. Dropping `take(7)` is a one-line deletion and a no-op under the default 7-day range (one entry per day is the store's invariant), so behaviour only changes for someone who has deliberately widened the range. Home now lists the window, like the Weight tab does.
- **Home gets no range control.** The issue notes the honest fixes are a Home-side control or a return to "the 7 most recent entries" — both Home-redesign calls. This PR does only what the issue asks for: name the range rather than misreport it.
- **Distinguishing the two empty cases through the repository interface**, via the new `isEmpty`, rather than `getOldestWeightDate()`, which #103 proposes deleting.
- Scope kept off #103's toes: the two streak banners are untouched here.

## Verification

Flutter is not installed in this environment and CLAUDE.md says not to install it, so `flutter analyze` / `flutter test` were **not** run locally — the `flutter_ci.yml` checks on this PR are the verification.

New tests in `test/ui/pages/home_page_test.dart` (in-memory repository, mirroring `weight_page_test.dart`) cover: the heading for each preset range, the onboarding prompt on an empty store, the range-named empty state when entries sit outside the window, and the history listing every entry in the selected range once the cap is gone. `weight_repository_test.dart` covers `isEmpty` across save, delete and `clear()`. Together with the existing `weight_page_test.dart` expectations, they pin every user-visible string across the getter consolidation.

Closes #99

### DIFF
--- a/lib/features/weight/data/weight_manager.dart
+++ b/lib/features/weight/data/weight_manager.dart
@@ -42,7 +42,7 @@ class WeightManager extends ChangeNotifier {
       .toList(growable: false);
 
   /// Whether the store holds any weigh-in at all, in or out of [historyRange].
-  bool get hasAnyWeights => _weightRepository.getAllWeights().isNotEmpty;
+  bool get hasAnyWeights => !_weightRepository.isEmpty;
 
   void selectHistoryRange(DateRange range) {
     if (range == _historyRange) return;
--- a/lib/features/weight/data/weight_manager.dart
+++ b/lib/features/weight/data/weight_manager.dart
@@ -41,6 +41,9 @@ class WeightManager extends ChangeNotifier {
       .where((weight) => _historyRange.contains(weight.date))
       .toList(growable: false);
 
+  /// Whether the store holds any weigh-in at all, in or out of [historyRange].
+  bool get hasAnyWeights => _weightRepository.getAllWeights().isNotEmpty;
+
   void selectHistoryRange(DateRange range) {
     if (range == _historyRange) return;
     _historyRange = range;
--- a/lib/ui/widgets/history_range_selector.dart
+++ b/lib/ui/widgets/history_range_selector.dart
@@ -10,16 +10,11 @@ const historyRangePresets = <DateRange>[
   DateRange.lastDays(365),
 ];
 
-extension HistoryRangeLabel on DateRange {
-  /// Rounded to the unit the span is closest to, so 180 days reads as months.
-  String get label => switch (days) {
-    180 => 'Last 6 months',
-    365 => 'Last year',
-    _ => 'Last $days days',
-  };
-
-  /// The span title-cased for a heading, as in 'Latest 30 Days of Weight'.
-  String get headingSpan => switch (days) {
+extension HistoryRangeSpan on DateRange {
+  /// Title-cased for headings ('Latest 30 Days of Weight'); lowercase it for
+  /// prose ('the last 30 days'). Rounded to the unit the span is closest to,
+  /// so 180 days reads as months.
+  String get span => switch (days) {
     180 => '6 Months',
     365 => 'Year',
     _ => '$days Days',
@@ -47,7 +42,10 @@ class HistoryRangeSelector extends StatelessWidget {
       onSelected: onSelected,
       itemBuilder: (context) => [
         for (final range in historyRangePresets)
-          PopupMenuItem(value: range, child: Text(range.label)),
+          PopupMenuItem(
+            value: range,
+            child: Text('Last ${range.span.toLowerCase()}'),
+          ),
       ],
       child: Padding(
         padding: const EdgeInsets.symmetric(vertical: 8.0),
@@ -55,7 +53,7 @@ class HistoryRangeSelector extends StatelessWidget {
           mainAxisSize: MainAxisSize.min,
           children: [
             Text(
-              selected.label,
+              'Last ${selected.span.toLowerCase()}',
               style: theme.textTheme.labelLarge?.copyWith(
                 color: theme.colorScheme.primary,
               ),
--- a/lib/ui/widgets/history_range_selector.dart
+++ b/lib/ui/widgets/history_range_selector.dart
@@ -17,6 +17,13 @@ extension HistoryRangeLabel on DateRange {
     365 => 'Last year',
     _ => 'Last $days days',
   };
+
+  /// The span title-cased for a heading, as in 'Latest 30 Days of Weight'.
+  String get headingSpan => switch (days) {
+    180 => '6 Months',
+    365 => 'Year',
+    _ => '$days Days',
+  };
 }
 
 /// Picks how far back the weight history reaches.
--- a/lib/features/weight/data/in_memory_weight_repository.dart
+++ b/lib/features/weight/data/in_memory_weight_repository.dart
@@ -31,4 +31,7 @@ class InMemoryWeightRepository with WeightRepositoryHelper implements WeightRepo
     invalidateCache();
     _weights.clear();
   }
+
+  @override
+  bool get isEmpty => _weights.isEmpty;
 }
--- a/lib/features/weight/data/persistent_weight_repository.dart
+++ b/lib/features/weight/data/persistent_weight_repository.dart
@@ -34,4 +34,7 @@ class PersistentWeightRepository with WeightRepositoryHelper implements WeightRe
     invalidateCache();
     await _box.clear();
   }
+
+  @override
+  bool get isEmpty => _box.isEmpty;
 }
--- a/lib/features/weight/data/weight_repository.dart
+++ b/lib/features/weight/data/weight_repository.dart
@@ -16,6 +16,10 @@ abstract class WeightRepository {
 
   Future<void> deleteWeight(DateTime date);
   Future<void> clear();
+
+  /// Whether the store holds no weigh-in. Implementations answer from the
+  /// store's own bookkeeping rather than by reading the entries.
+  bool get isEmpty;
   
   /// Returns the lowest weight recorded since [since] (inclusive).
   /// If [since] is null, returns the all-time lowest weight.
--- a/lib/ui/pages/home_page.dart
+++ b/lib/ui/pages/home_page.dart
@@ -62,7 +62,7 @@ class HomePage extends StatelessWidget {
                 mainAxisAlignment: MainAxisAlignment.spaceBetween,
                 children: [
                   Text(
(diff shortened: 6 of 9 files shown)


igorbasko01:
**Verdict: nothing blocks merge.** The change is small, correct, matches what #99 asked for, and is covered by tests. Two non-blocking notes below.

## Invariants checked

Nothing here goes near the risky surfaces, and I verified that rather than assuming it:

- **Generated code / Hive typeIds / Drift schema** — no model, table, `*.g.dart` or `hive_registrar.g.dart` touched. `schemaVersion` untouched, correctly.
- **Day granularity** — no new `DateTime` is stored or compared. `hasAnyWeights` is a pure read, and the test helper `daysAgo` builds `DateTime(y, m, d - n)`, already normalised.
- **Mutation shape** — no new mutation path; `hasAnyWeights` adds a read, not a write, so the write-through-repository-then-reload contract is untouched.
- **Interfaces over stores** — `hasAnyWeights` goes through `WeightRepository.getAllWeights()`, not a Hive box. No new `MultiProvider` wiring needed.
- **Backup compatibility** — no codec, CSV column, or `SerializationService` change.
- **Versioning** — `pubspec.yaml` and `CHANGELOG.md` untouched; title carries a `fix:` prefix that matches the change.
- **Tests** — `test/ui/pages/home_page_test.dart` mirrors `lib/ui/pages/home_page.dart` and uses `InMemoryWeightRepository`, closely following `weight_page_test.dart` (including the viewport bump, which is genuinely needed here — the header plus both streak banners push the list below an 800x600 fold).

## Correctness pass

The two things I most expected to be wrong are both fine:

- **The diff column survives the cap removal.** In the old code `latest7[index]` was `history[index]` by construction, so replacing `latest7` with `history` in `_buildHistorySection` leaves `item.value - history[index + 1].value` computing exactly the same thing. `getWeightsSince` sorts descending (`weight_repository.dart:94`), so the "sorted descending" comment on that block still holds for the widened list, and the oldest in-range entry correctly gets a `null` diff — same as the 7th row did before.
- **The two empty states can't be confused.** `_weights` is loaded via `getWeightsSince(_historyRange.from)`, so it genuinely cannot see entries below the window — `hasAnyWeights` reading `getAllWeights()` directly is the right call, not redundancy. The one case that could have flipped the branch, a store holding only future-dated entries, doesn't: `DateRange.contains` is unbounded above when `to == null`, so those land in `history` and the empty state never fires.

The extra `getAllWeights()` per Home build is genuinely marginal, as the PR body claims — `calculateOvereatingStats` already calls `getOldestWeightDate()`, which calls `getAllWeights()` on the same build.

## Notes (neither blocks merge)

1. **`headingSpan` and `label` are two switches over `days` that must now stay in sync** — `lib/ui/widgets/history_range_selector.dart:13-26`. Today they agree on every entry in `historyRangePresets`, and the fallback arms agree for arbitrary day counts too, so there is no current bug. The hazard is future: special-casing a preset in one switch and not the other (say `30 => 'Last month'` in `label`) puts "Latest 30 Days of Weight" above a selector reading "Last month". Worth a glance if a preset is ever reworded.

2. **The squashed changelog line will undersell the change.** "name the selected history range on Home" doesn't hint that Home now lists every entry in the window rather than at most 7 rows. `fix:` is the right prefix — the cap removal is part of making the heading honest, not a separate feature — but a `BREAKING CHANGE`-free extra line in the commit body, or a slightly wider subject, would leave a better changelog entry for anyone who has widened their range.

Scope discipline is good: the streak banners are left alone for #103, and `getOldestWeightDate()` is deliberately not leaned on.

---

lib/ui/widgets/history_range_selector.dart
igorbasko01: I'm not sure we need the word 'last' as part of the translation. And then we maybe can consolidate the label and the span. Wdyt?

igorbasko01: Agreed — done in 11a107b. There's now one getter, `span`, title-cased and without the leading "Last":

```dart
extension HistoryRangeSpan on DateRange {
  String get span => switch (days) {
    180 => '6 Months',
    365 => 'Year',
    _ => '$days Days',
  };
}
```

Each call site supplies its own framing word: `'Last ${range.span.toLowerCase()}'` for the selector and the two empty states, `'Latest ${range.span} of Weight'` for the Home heading. Title case is the direction that survives a round trip — `'6 Months'.toLowerCase()` gives back exactly `'6 months'`, whereas going the other way would need capitalising per word.

Every rendered string is byte-identical to before, which the existing tests pin: `weight_page_test.dart` still finds `'Last 7 days'` and `'No weight entries in the last 7 days. Tap + to log your weight.'`, and `home_page_test.dart` still finds `'Latest 6 Months of Weight'`.

This also closes the desync risk the review comment above flagged — there's only one switch over `days` now.

---

lib/features/weight/data/weight_manager.dart
```diff
@@ -41,6 +41,9 @@ class WeightManager extends ChangeNotifier {
       .where((weight) => _historyRange.contains(weight.date))
       .toList(growable: false);
 
+  /// Whether the store holds any weigh-in at all, in or out of [historyRange].
+  bool get hasAnyWeights => _weightRepository.getAllWeights().isNotEmpty;
```
igorbasko01: Is there a cheaper way to test if a box is empty instead of pulling the whole data and then check if it is empty?

igorbasko01: Yes — fixed in b12b2e5. The store now answers from its own bookkeeping instead of materialising every entry:

```dart
// WeightRepository
/// Whether the store holds no weigh-in. Implementations answer from the
/// store's own bookkeeping rather than by reading the entries.
bool get isEmpty;

// PersistentWeightRepository
bool get isEmpty => _box.isEmpty;   // BoxBase.isEmpty → length == 0, no deserialization

// InMemoryWeightRepository
bool get isEmpty => _weights.isEmpty;

// WeightManager
bool get hasAnyWeights => !_weightRepository.isEmpty;
```

`Box.values.toList()` was deserialising the whole box on every Home rebuild just to look at `.isEmpty`; `BoxBase.isEmpty` is a key-count check, so it's O(1) and reads nothing.

I put it on the interface rather than defaulting it on `WeightRepositoryHelper`, so an implementation can't inherit the expensive version by accident — the doc comment states the contract. Both existing implementations are the only ones (`PersistentWeightRepository` and `InMemoryWeightRepository`), and `weight_repository_test.dart` covers the transitions: empty at the start, false after a save, true again after a delete and after `clear()`.

---

PR #111, opened 2026-08-27
feat: clear all stored data from Settings (#101)

### LINKED ISSUE #101: feat: add "Clear All Data" button in Settings with a confirmation modal
## Summary

The Settings tab currently offers **Export Data** and **Import Data**, but there is no way to wipe the app's local data. Users who want a clean slate (new phone, restarting a tracking cycle, testing) have to uninstall and reinstall the app.

Add a **Clear All Data** action to the Settings page that erases both stores, guarded by a confirmation modal so it can't be triggered by a stray tap.

### Placement

Put it in its own section at the **bottom of the page**, separate from the "Data Management" card — not as a third row inside it. Export and Import read as a group of reversible operations; a destructive row directly below Import is adjacent to the actions users tap routinely, which is where a mis-tap costs the most. A separate trailing section also matches the usual "danger zone" convention.

The page is a single `ListView` with one `_buildHeader(theme, 'Data Management')` section, so this is an extension of the existing structure:

- a second header below the existing card, e.g. "Danger Zone"
- its own `Card`, styled like the first but with the error color for the border/tint so it reads as distinct
- a single `ListTile`:
  - leading icon: `Icons.delete_forever`, tinted with `theme.colorScheme.error`
  - title: "Clear All Data"
  - subtitle: something like "Permanently delete all weight and bite data."
  - `enabled: !_busy`, matching the export/import tiles

### Confirmation modal

Tapping it opens an `AlertDialog`:

- states that the action is permanent and cannot be undone, and that it removes weight history, the bite log and pacing config
- suggests exporting a backup first
- **Cancel** (default) and a destructive **Clear** action

### On confirm

- clear all three datasets — `WeightRepository.clear()`, `BiteRepository.clearBites()`, `BiteRepository.clearPacingConfigs()`
- refresh the managers so the UI updates immediately (`WeightManager` reload, `BiteManager.refresh()`), same as the post-restore path
- show the progress snackbar while clearing (reuse `_progressSnackBar`) and a success/failure snackbar afterwards, mirroring `_exportData` / `_importData`

Pacing config: after clearing, the default config should be re-seeded on next read so the Bite tab keeps working (same "seeded with a default on first run" behaviour as a fresh install) — worth verifying which path re-seeds it.

## Implementation notes

The destructive core already exists in `SerializationService.restoreFromBackup` (`lib/features/settings/data/serialization_service.dart`), which does clear-then-restore. Extracting the clear half into a testable `clearAllData()` on `SerializationService` would keep the file-picker/IO separation the class already follows, and lets the new path be unit-tested without touching the UI.

## Acceptance criteria

- [ ] "Clear All Data" lives in its own section at the bottom of Settings, below the Data Management card, styled as destructive
- [ ] Tapping it always shows a confirmation modal; dismissing/cancelling changes nothing
- [ ] Confirming removes all weights, all bites and pacing config history
- [ ] Weight and Bite tabs reflect the empty state without an app restart
- [ ] The Bite tab still works after clearing (pacing config falls back to the default)
- [ ] Unit test covering the clear path (in-memory weight repo + `BiteDatabase.forTesting`)
- [ ] `flutter analyze` and `flutter test` pass

igorbasko01:
## What changed

Settings could export and import a backup but never wipe the app's data, so a clean slate meant uninstalling and reinstalling. This adds a **Clear All Data** action.

- **`lib/ui/pages/settings_page.dart`** — a second section, "Danger Zone", below the Data Management card: its own error-tinted `Card` with a single `ListTile` (`Icons.delete_forever`, error-tinted, `enabled: !_busy`). Tapping it raises an `AlertDialog` — "Clear all data?" / Cancel / a destructive **Clear** — that says the action is permanent, names what goes (weight history, bite log, pacing settings), and suggests exporting first. On confirm it shows the existing `_progressSnackBar`, then a success or failure toast, mirroring `_exportData` / `_importData`.
- **`lib/features/settings/data/serialization_service.dart`** — `clearAllData(weightRepo, biteRepo)`: `WeightRepository.clear()`, `BiteRepository.clearBites()`, `BiteRepository.clearPacingConfigs()`, then re-seeds the default pacing config.
- **`lib/features/bite/data/bite_database.dart`** — the starting thresholds move into a top-level `defaultPacingConfig` const, which `_seedDefaultPacingConfig` now inserts, so the seed and the re-seed can't drift apart.
- **`CLAUDE.md`** — one clause noting `clearAllData` alongside `restoreFromBackup`, so the architecture note stays accurate.

## Acceptance criteria

- [x] **Own section at the bottom of Settings, below the Data Management card, styled as destructive** — `_buildHeader(theme, 'Danger Zone')` after a `SizedBox(height: 24)`, then a `Card` tinted with `errorContainer` and bordered with `error`.
- [x] **Tapping always shows a confirmation modal; dismissing/cancelling changes nothing** — `_confirmClear()` runs before any repository call and returns `false` on Cancel or a barrier dismissal. Covered by *asks before deleting anything* and *cancelling leaves the stores untouched*.
- [x] **Confirming removes all weights, all bites and pacing config history** — covered at the service level with a recording repository (asserting the exact mutation sequence) and against a real store (`BiteDatabase.forTesting(NativeDatabase.memory())`).
- [x] **Weight and Bite tabs reflect the empty state without an app restart** — `weightManager.refresh()` / `biteManager.refresh()` after the clear, exactly as the post-restore path does. The widget test asserts `weightManager.history` empties and `biteManager.todayCount` drops to 0.
- [x] **The Bite tab still works after clearing (pacing config falls back to the default)** — see the decision below; the widget test asserts `biteManager.b2` returns from a retuned 25 s to the default 30 s, and the Drift test asserts `pacingConfigAt` resolves to the default with exactly one version stored.
- [x] **Unit test covering the clear path (in-memory weight repo + `BiteDatabase.forTesting`)** — added to `test/features/settings/data/serialization_service_test.dart`, plus five widget tests in `test/ui/pages/settings_page_test.dart`.
- [x] **`flutter analyze` and `flutter test` pass** — green on the `analyze_and_test` CI job for this head, see *Verification*.

## Decisions made

**Re-seeding the pacing config.** The issue flagged this as worth verifying, and the answer is that nothing re-seeds it in-session: `_seedDefaultPacingConfig` runs only from the database's `beforeOpen`, so after `clearPacingConfigs()` the store has no config until the next app start. `BiteManager.refresh()` keeps whatever version it already held in that case (`pacingConfigAt(now) ?? _pacingConfig`), so thresholds a restore had replaced would survive a clear that was supposed to delete them. `clearAllData` therefore writes the default back through `setPacingConfig`, which leaves the store in exactly a fresh install's state — one version, effective from the epoch. Changing `clearPacingConfigs` itself to re-seed was rejected: `restoreFromBackup` clears then inserts the backup's own versions, and a seeded row would linger in that history.

**Where the default lives.** It was duplicated as a literal in `_seedDefaultPacingConfig` and as `BiteManager`'s fallback constants. Rather than adding a third copy, the seed now reads a `defaultPacingConfig` const in `bite_database.dart` and `clearAllData` writes that same value. `BiteManager`'s private fallbacks were left alone — changing them is outside this issue.

**Signature of `clearAllData`.** It takes the two repositories directly, like `restoreFromBackup`, rather than a `BuildContext` like `exportData` / `importData`. That keeps the destructive path unit-testable with no widget tree, at the cost of the page reading `WeightRepository` and `BiteRepository` from `Provider` — both are already wired in `lib/main.dart`.

**Confirm before the busy state.** `_clearAllData` reads its providers up front, then asks; `_busy` and the progress toast only start once the user has confirmed, so a cancelled clear leaves no trace on screen. This matches the import path's behaviour with a dismissed picker.

**Wording.** The tile subtitle is the issue's suggested "Permanently delete all weight and bite data."; the dialog spells out the pacing settings as well, since they are cleared too.

## Verification

Flutter is not installed in this environment and CLAUDE.md says not to install it here, so `flutter analyze` and `flutter test` were **not** run locally — the `flutter_ci.yml` checks on this PR are the verification, and `analyze_and_test` passed on this head. The diff was kept small and self-reviewed for that reason. No generated code changed: no Hive model, `typeId`, or Drift table was touched, so no `build_runner` run and no `schemaVersion` bump are needed.

Closes #101


igorbasko01:
**Verdict: nothing blocks merge.** The change is small, correctly layered, and the two things this repo is easiest to get wrong here — the pacing-config re-seed and the manager refresh — are both handled and tested. Two non-blocking notes below.

## Invariant checks

All clean; recording them so the next reader doesn't have to re-derive them:

- **Generated code** — no `@HiveType`/`@HiveField` or `@DriftDatabase` table touched, so no stale `*.g.dart` and no `build_runner` run needed. `schemaVersion` correctly left at 2.
- **Interfaces over stores** — `clearAllData` takes `WeightRepository`/`BiteRepository`, and `settings_page.dart` reads both from `Provider`. Both are already registered in the `MultiProvider` in `lib/main.dart:38-39`, so the new `context.read` calls resolve at runtime; no wiring gap.
- **Mutation shape** — writes go through the repositories, then `weightManager.refresh()` / `biteManager.refresh()` re-read, matching the post-restore path added in #109.
- **Backup compatibility** — no codec or CSV column changed; `restoreFromBackup`'s null-vs-empty tolerance is untouched.
- **Derived, not stored** — nothing new is persisted; the re-seeded config is a real stored version, which is exactly what a fresh install has.
- **Cache invalidation** — `PersistentWeightRepository.clear()` calls `invalidateCache()`, so the lowest-weight cache in `WeightRepositoryHelper` can't survive the clear and report a lowest weight for an empty store.
- **Versioning** — `feat:` is the right prefix for a user-visible feature, and neither `pubspec.yaml` nor `CHANGELOG.md` is touched.
- **Tests** — mirror `lib/`, and the Drift-backed case uses `BiteDatabase.forTesting(NativeDatabase.memory())` as CLAUDE.md asks. The reseed assertion (`allPacingConfigs()` has length 1, `pacingConfigAt` resolves to the default) is the right shape: it pins "a fresh install's state", not just "not null".

The re-seed decision is the correct one, and for the reason the description gives: `_seedDefaultPacingConfig` only runs from `beforeOpen`, and `BiteManager.refresh()` falls back with `?? _pacingConfig`, so without the explicit `setPacingConfig` a clear would leave retuned thresholds alive in memory after the store said they were gone.

## Findings

**1. A partial failure leaves the Weight tab showing data that is already gone.** `lib/ui/pages/settings_page.dart:194-207`

`clearAllData` is four sequential awaits with no transaction, and the `catch` shows the failure toast without refreshing either manager. If `weightRepo.clear()` succeeds and `biteRepo.clearBites()` then throws (SQLite I/O error, disk full, database locked), the weights are gone from Hive but `WeightManager._weights` still holds them. That matters more for weight than for bite: `bite_page.dart:79` refreshes `BiteManager` whenever the tab becomes visible, but nothing refreshes `WeightManager` outside these two Settings paths — so the Weight tab keeps listing weigh-ins that no longer exist until an app restart or the next weight mutation. Refreshing both managers in the `finally` (guarded on `mounted`) rather than only on the success path would close it. `_importData` has the same shape, so this is a pre-existing pattern being followed rather than something the PR introduces.

**2. Optional: `BiteManager`'s fallback thresholds are now a third copy of 15/30 s.** `lib/features/bite/data/bite_manager.dart:40-41`

The description explicitly scopes this out, which is fair. But now that `defaultPacingConfig` exists as the canonical value, `_fallbackB1`/`_fallbackB2` are the only remaining place the numbers are spelled out independently — and they're the copy that would silently disagree if the default ever moved. A one-line change to derive them from `defaultPacingConfig` would finish the deduplication the PR started. Not worth blocking on.

---

igorbasko01:
Acted on the review above.

**Finding 1 (no transaction / no refresh on failure)** — fixed in 8377ca9. `_clearAllData` now refreshes `WeightManager` and `BiteManager` in a `finally` rather than only on success, so a clear that throws after the weights are gone can't leave the Weight tab listing weigh-ins that no longer exist. Added a widget test that fails the clear after the weights have been emptied and asserts the manager is empty behind the failure toast. Left the pre-existing `_importData` alone — same shape, but outside this PR's scope.

**Finding 2 (`_fallbackB1`/`_fallbackB2` vs `defaultPacingConfig`)** — left as is, deliberately. Folding `BiteManager`'s fallbacks onto the new const is a sensible follow-up but touches the pacing path, which #101 doesn't cover.

---

igorbasko01:
Run test apk build on this branch

igorbasko01:
Test APK built and published: [run #8](https://github.com/igorbasko01/food-locker/actions/runs/33141696806) — succeeded, all steps green.

- **Build:** `1.17.0-pr111.8`, from this branch's head `8377ca9` (both commits included)
- **Install:** https://github.com/igorbasko01/food-locker/releases/download/test-builds/food_locker-test.apk

Installs alongside the production app as **FoodLocker Test** (`com.baskorp.food_locker.test`) with its own empty data. As the release notes say, this replaced the rolling `test-builds` prerelease, so the download URL is unchanged.

Worth exercising on device, since the widget tests can't cover it: Settings → the new **Danger Zone** section sitting under Data Management, the confirmation dialog, and — after confirming — that the Weight and Bite tabs are empty without a restart and the Bite tab still paces normally on the default 15 s / 30 s thresholds.

---

PR #112, opened 2026-08-28
feat: remove the Home streak banners and overeating analytics

igorbasko01:
## What changed

Removes both Home banners and the whole overeating/streak calculation they were the only consumers of, down to the repository method that existed only to feed it.

**Deleted**
- `lib/ui/widgets/streak_banner.dart`
- `lib/ui/widgets/longest_streak_banner.dart`
- `lib/ui/widgets/day_date_text.dart`
- `test/ui/widgets/streak_banner_test.dart`
- `test/ui/widgets/longest_streak_banner_test.dart`

**Edited**
- `lib/ui/pages/home_page.dart` — both banner imports, the `overeatingStats` local, the sliver wrapping the banners, and the tagline (see decisions).
- `lib/features/weight/data/weight_analytics.dart` — `StreakType`, `OvereatingStats`, `_isOvereaten`, `calculateOvereatingStats`, `_calculateCurrentStreak`, `_calculateLongestCleanStreak`, and the now-unused `DateTimeCalendar` extension. The three `lowest*` getters stay.
- `lib/features/weight/data/weight_manager.dart` — the `show StreakType, OvereatingStats` export and the `overeatingStats` getter.
- `lib/features/weight/data/weight_repository.dart` — `getOldestWeightDate()` from the interface and its `WeightRepositoryHelper` body.
- `test/features/weight/weight_manager_test.dart` — the `WeightManager Overeating & Streak Stats` group.
- `test/features/weight/weight_repository_test.dart` — the `getOldestWeightDate` test.
- `test/ui/pages/home_page_test.dart` — a comment that explained the oversized test surface in terms of the banners this PR deletes.
- `README.md`, `CLAUDE.md`, `lib/features/bite/data/bite_analytics.dart` doc comment.

Net: 9 insertions, 619 deletions across 15 files.

## Acceptance criteria

- [x] **Neither banner renders on Home; the page shows header → history section → version label** — the banner `SliverToBoxAdapter` is gone; the slivers are now header → history heading → history list → `AppVersionLabel`.
- [x] **Both banner widgets and their tests are deleted** — plus `day_date_text.dart`, see decisions below.
- [x] **No `StreakType`, `OvereatingStats`, `overeatingStats`, or streak calculation remains in `lib/`** — verified by a case-insensitive repo-wide grep for `overeat|streak|DateTimeCalendar|DayDateText` over `lib/` and `test/`: zero hits. `WeightAnalytics` is now 17 lines.
[...]

PR #114, opened 2026-08-29
feat: add a 52-week weight-change heatmap to Home

### LINKED ISSUE #105: feat: 52-week calendar heatmap on Home screen for weekly weight change
## Summary

Add a calendar heatmap to the Home screen: a grid of 52 cells, each representing one calendar week over the past year. Each cell is colored by whether that week&#39;s weight went **up** or **down**, with color **intensity scaled to the magnitude of the delta**.

- 🟥 **Red** — gained weight that week
- 🟩 **Green** — lost weight that week
- Darker / more saturated = larger change

This gives a glanceable, at-a-glance view of trend direction across the year that the existing line chart / analytics don&#39;t surface in one look.

## Motivation

The weight feature already tracks day-granular entries and computes streaks / overeating / lowest-weight analytics, but there&#39;s no compact &#34;how has each week gone?&#34; overview. A GitHub-contributions-style heatmap makes long-run momentum obvious.

## Design decisions (resolved)

All open questions were resolved — see [this comment](https://github.com/igorbasko01/food-locker/issues/105#issuecomment-5460716400) for the rationale behind each.

**Week delta** — Intra-week only: last entry of the week − first entry of the same week. A week needs **≥ 4 entries** to produce a delta; fewer than 4 → no data. No carry-forward across week boundaries, so every cell is self-contained.

**Gray cells** — No dead zone. Gray means exactly one thing: fewer than 4 entries that week. Any week with ≥ 4 entries is colored; an exact 0.0 delta lands in the faintest **green** bucket. (The in-progress current week is therefore gray until it reaches 4 entries.)

**Boundary & layout** — Weeks run **Sunday → Saturday**. Grid is **4 rows × 13 columns** (52 cells), oldest top-left → newest bottom-right, last cell = the current week. No month labels.

**Intensity** — Fixed, unit-aware buckets, 4 levels per direction:

| Level | kg | lb |
|---|---|---|
| 1 (faintest) | < 0.25 | < 0.5 |
| 2 | 0.25 – 0.5 | 0.5 – 1.0 |
| 3 | 0.5 – 1.0 | 1.0 – 2.0 |
| 4 (strongest) | ≥ 1.0 | ≥ 2.0 |

Bucket set chosen from the week&#39;s entry `WeightUnit` (every stored entry is `kilograms` today, so the lb column is forward-looking).

**Color semantics** — loss = green / gain = red, hard-coded for now, matching the app&#39;s existing loss-oriented framing. The delta → color decision goes through a single mapping function so a future &#34;goal direction&#34; setting can flip it.

**Placement** — Top of Home, directly under the title block and above the streak banners.

**Interaction** — None in this iteration. No tap-to-drill-in.

**Accessibility** — Deferred: no legend, no lightness ramp, no texture or shape cue. ⚠️ Known gap — red/green is the only signal, which is the worst pairing for color-blind users. Track as a follow-up once the widget lands.

### Implementation notes

- Dart&#39;s `DateTime.weekday` is Mon=1…Sun=7 — the Sunday-start offset must be explicit (`weekday % 7`).
- Use calendar-component date math (as in the existing `DateTimeCalendar` extension) rather than `Duration(days:)`, to stay DST-safe.

## Scope notes

- Weight-only feature; lives under `lib/features/weight/` (analytics) + `lib/ui/` (widget), wired through the existing `WeightManager` / `WeightAnalytics` layering. The weekly-delta computation is pure and belongs in/near `WeightAnalytics` so it&#39;s unit-testable with `InMemoryWeightRepository`.
- No persistence/model changes expected — this is derived from existing stored weights, so no Hive `typeId` or migration work.

igorbasko01:
A GitHub-contributions-style grid on Home: 52 cells, one per calendar week of the past year, coloured by that week's weight change.

## What changed

| File | Role |
|---|---|
| `lib/features/weight/data/weekly_weight_change.dart` | New `WeeklyWeightChange` value type — a week's `weekStart`, `delta`, `unit`, plus `hasData` / `isGain` / `level`. |
| `lib/features/weight/data/weight_analytics.dart` | New `weeklyChanges({weeks, asOf})` — pure computation over the repository. |
| `lib/features/weight/data/weight_manager.dart` | `weeklyChanges` getter delegating to analytics. |
| `lib/ui/widgets/weekly_change_color.dart` | The single delta-to-colour mapping function. |
| `lib/ui/widgets/weekly_change_heatmap.dart` | The 4 × 13 grid widget. |
| `lib/ui/pages/home_page.dart` | Places the grid under the title block. |

No model, Hive `typeId`, Drift schema or migration work — the grid is derived entirely from stored weights at read time, as the issue anticipated.

## How each resolved decision is met

**Week delta — intra-week, ≥ 4 entries.** `WeightAnalytics.weeklyChanges` buckets weigh-ins by their week and reports `last − first` within that bucket only; a bucket under `WeeklyWeightChange.minEntries` (4) reports `delta == null`. Nothing crosses a week boundary. Covered by `never carries across a week boundary` and `splits a Saturday-to-Sunday run at the boundary`.

**Gray means one thing.** `delta == null` — fewer than 4 entries — is the only path to `weeklyChangeNoDataColor`. The in-progress current week is therefore gray until it reaches 4.

**Exact 0.0 → faintest green.** `isGain` is `delta > 0`, so a flat week takes the loss ramp; `level` buckets on `abs(delta)`, putting 0.0 at level 1. Covered by `an exact zero is the faintest loss, never a gain`.

**Sunday → Saturday, 4 × 13, oldest top-left.** `_weekStart` uses the explicit `weekday % 7` offset the issue called for. The widget fills row-major from a list the analytics returns oldest-first, so the last cell is the current week. Covered by `every week start is a Sunday` and `the current week takes the last cell`.

**Fixed, unit-aware buckets.** `WeeklyWeightChange.level` returns 1–4 against `< 0.25 / 0.5 / 1.0` kg, doubled for `WeightUnit.pounds` (`< 0.5 / 1.0 / 2.0`). The bucket set comes from the week's last weigh-in's unit. Both columns of the issue's table are covered by `kilogram buckets run 0.25 / 0.5 / 1.0` and `pound buckets are the kilogram ones doubled`.

**Loss = green / gain = red through one function.** `weeklyChangeColor` is the only place the sign becomes a colour; swapping `_lossRamp` and `_gainRamp` there is the whole of a future goal-direction setting. The widget never sees a sign.

**Placement.** Top of Home, directly under the title block.

**No interaction, no legend, no month labels.** The grid is a plain `DecoratedBox` per cell.

**DST safety.** Both `_weekStart` and the walk back through the year use calendar-component arithmetic (`DateTime(y, m, d - n)`), never `Duration(days:)`, as the issue's implementation note asked.

- **The issue says "above the streak banners"; those are gone.** #112 removed them from Home before this landed. The grid goes directly under the title block, which is what that instruction resolves to now.
- **`DateTimeCalendar` no longer exists.** #112 removed it along with the overeating analytics. Rather than reviving an extension for two call sites, the date math is written inline as `DateTime(y, m, d - n)` — the same calendar-component form, matching what `DateRange` in `lib/core/` already does.
- **The grid hides itself while no week has data.** Unspecified in the issue. A wall of 52 gray cells above the "No weight entries yet" onboarding text reads as broken rather than empty, and Home already suppresses its history section the same way. One `if` in `HomePage`; the widget itself renders unconditionally. Easy to drop if you'd rather always show the frame.
- **A week's unit comes from its *last* weigh-in.** The issue says "the week's entry `WeightUnit`" without picking one for a mixed week. The last entry is the one the delta is measured to, and every stored entry is `kilograms` today regardless.
- **`weeklyChanges` takes an `asOf` parameter.** Injecting "now" is what makes the week-boundary and grid-extent tests deterministic; production callers use the default.
- **The widget tolerates a short list.** Cells past the end of `weeks` draw as no-data, so the 4 × 13 frame is never partially built.

## Verification

**Flutter is not installed in this environment, and per `CLAUDE.md` it was not installed** — so `flutter analyze` and `flutter test` were never run locally. The `flutter_ci.yml` check on this PR is the verification, and `analyze_and_test` is **green on `978ba24`**. The diff is additive (no existing behaviour altered) and was self-reviewed before each push.

Tests added:

- `test/features/weight/weekly_weight_change_test.dart` — 21 cases over `InMemoryWeightRepository`: grid extent and ordering, Sunday starts, the 4-entry threshold at 3 and at 4, boundary non-carry, out-of-grid and future-dated entries, chronological rather than insertion order, unit selection, and every bucket edge in both kg and lb.
- `test/ui/widgets/weekly_change_heatmap_test.dart` — 52 cells in 4 rows, no-data colouring, gain vs. loss distinguishable, magnitude darkening, current week last, short-list tolerance.
- `test/ui/pages/home_page_test.dart` — the grid appears under the title block and above the history header once a week has data, and stays hidden when none does.

A second commit (`978ba24`) responds to the independent review below: it pins the grid shape to `WeightAnalytics.heatmapWeeks` rather than a hardcoded 52, and asserts `weeks > 0`. See [that comment](https://github.com/igorbasko01/food-locker/pull/114#issuecomment-5465466003) for what was fixed and what was deliberately left as follow-up.

## Known gaps — follow-up material, not blockers

- **Accessibility, as the issue accepted.** Red/green is the only signal, with no legend, lightness ramp, or texture cue.
- **Mixed-unit weeks.** A week holding both kg and lb entries yields an unconverted delta. Latent only — `AddWeightDialog` hard-codes kilograms; a hand-edited backup is the sole route in.
- **`weeklyChanges` recomputes on every `HomePage` build.** Same shape as the pagination `TODO` already on `WeightManager._weights`.

Closes #105

### DIFF
--- a/lib/features/weight/data/weight_analytics.dart
+++ b/lib/features/weight/data/weight_analytics.dart
@@ -1,9 +1,14 @@
+import 'package:food_locker/features/weight/data/weekly_weight_change.dart';
+import 'package:food_locker/features/weight/data/weight.dart';
 import 'package:food_locker/features/weight/data/weight_repository.dart';
 
 class WeightAnalytics {
   final WeightRepository _weightRepository;
 
   WeightAnalytics(this._weightRepository);
+
+  /// Weeks the Home heatmap draws.
+  static const int heatmapWeeks = 52;
 
   double? get lowestAllTime => _weightRepository.getLowestWeight();
 
@@ -14,4 +19,61 @@ class WeightAnalytics {
   double? get lowestLast7Days => _weightRepository.getLowestWeight(
     since: DateTime.now().subtract(const Duration(days: 7)),
   );
+
+  /// The last [weeks] Sunday-to-Saturday weeks of weight change, oldest first
+  /// and ending with the week holding [asOf] — the week still in progress.
+  ///
+  /// Always [weeks] long: a week the store has too little for is present with
+  /// no delta rather than missing, so callers can index it as a fixed grid.
+  List<WeeklyWeightChange> weeklyChanges({
+    int weeks = heatmapWeeks,
+    DateTime? asOf,
+  }) {
+    final weekStarts = _weekStartsEndingAt(asOf ?? DateTime.now(), weeks);
+    final entriesByWeek = <DateTime, List<Weight>>{};
+
+    for (final weight in _weightRepository.getWeightsSince(weekStarts.first)) {
+      final start = _weekStart(weight.date);
+      // getWeightsSince is unbounded above, so future-dated weigh-ins arrive
+      // with no cell to sit in.
+      if (start.isAfter(weekStarts.last)) continue;
+      entriesByWeek.putIfAbsent(start, () => []).add(weight);
+    }
+
+    return [
+      for (final start in weekStarts) _change(start, entriesByWeek[start]),
+    ];
+  }
+
+  static WeeklyWeightChange _change(DateTime weekStart, List<Weight>? entries) {
+    if (entries == null || entries.length < WeeklyWeightChange.minEntries) {
+      return WeeklyWeightChange(weekStart: weekStart);
+    }
+
+    entries.sort((a, b) => a.date.compareTo(b.date));
+    final first = entries.first;
+    final last = entries.last;
+    return WeeklyWeightChange(
+      weekStart: weekStart,
+      delta: last.value - first.value,
+      unit: last.unit,
+    );
+  }
+
+  /// The Sunday opening [date]'s week, at local midnight. Dart counts Mon=1…
+  /// Sun=7, so `% 7` maps Sunday to a zero offset.
+  static DateTime _weekStart(DateTime date) =>
+      DateTime(date.year, date.month, date.day - (date.weekday % 7));
+
+  /// [weeks] consecutive week starts, oldest first, the last being [asOf]'s.
+  ///
+  /// Stepping by calendar components rather than `Duration(days: 7)` keeps the
+  /// walk on Sundays across a daylight-saving shift.
+  static List<DateTime> _weekStartsEndingAt(DateTime asOf, int weeks) {
+    final current = _weekStart(asOf);
+    return [
+      for (var back = weeks - 1; back >= 0; back--)
+        DateTime(current.year, current.month, current.day - 7 * back),
+    ];
+  }
 }
--- a/lib/features/weight/data/weight_analytics.dart
+++ b/lib/features/weight/data/weight_analytics.dart
@@ -29,6 +29,7 @@ class WeightAnalytics {
     int weeks = heatmapWeeks,
     DateTime? asOf,
   }) {
+    assert(weeks > 0);
     final weekStarts = _weekStartsEndingAt(asOf ?? DateTime.now(), weeks);
     final entriesByWeek = <DateTime, List<Weight>>{};
 
--- a/lib/features/weight/data/weekly_weight_change.dart
+++ b/lib/features/weight/data/weekly_weight_change.dart
@@ -1,0 +1,53 @@
+import 'package:food_locker/features/weight/data/weight.dart';
+
+/// One calendar week's weight change, as the Home heatmap reads it.
+///
+/// Intra-week only: [delta] is the week's last weigh-in minus its first, so
+/// nothing carries across a week boundary and every week stands on its own. A
+/// week under [minEntries] weigh-ins reports no [delta] at all.
+class WeeklyWeightChange {
+  const WeeklyWeightChange({required this.weekStart, this.delta, this.unit});
+
+  /// Weigh-ins a week needs before it reports a [delta].
+  static const int minEntries = 4;
+
+  /// The Sunday opening the week, at local midnight.
+  final DateTime weekStart;
+
+  /// Last weigh-in of the week minus its first, or null under [minEntries].
+  final double? delta;
+
+  /// The unit [delta] is expressed in; null exactly when [delta] is.
+  final WeightUnit? unit;
+
+  bool get hasData => delta != null;
+
+  /// Whether the week ended heavier. A flat week counts as not gained.
+  bool get isGain => (delta ?? 0) > 0;
+
+  /// Magnitude bucket, 1 (faintest) to 4 (strongest), or null without a
+  /// [delta]. The pound bounds are the kilogram ones doubled.
+  int? get level {
+    final magnitude = delta?.abs();
+    if (magnitude == null) return null;
+    final scale = unit == WeightUnit.pounds ? 2.0 : 1.0;
+    if (magnitude < 0.25 * scale) return 1;
+    if (magnitude < 0.5 * scale) return 2;
+    if (magnitude < 1.0 * scale) return 3;
+    return 4;
+  }
+
+  @override
+  bool operator ==(Object other) =>
+      other is WeeklyWeightChange &&
+      other.weekStart == weekStart &&
+      other.delta == delta &&
+      other.unit == unit;
+
+  @override
+  int get hashCode => Object.hash(weekStart, delta, unit);
+
+  @override
+  String toString() =>
+      'WeeklyWeightChange(weekStart: $weekStart, delta: $delta, unit: $unit)';
+}
--- a/lib/features/weight/data/weight_manager.dart
+++ b/lib/features/weight/data/weight_manager.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/foundation.dart';
 import 'package:food_locker/core/date_range.dart';
+import 'package:food_locker/features/weight/data/weekly_weight_change.dart';
 import 'package:food_locker/features/weight/data/weight.dart';
 import 'package:food_locker/features/weight/data/weight_analytics.dart';
 import 'package:food_locker/features/weight/data/weight_repository.dart';
@@ -95,4 +96,7 @@ class WeightManager extends ChangeNotifier {
   double? get lowestAllTime => _analytics.lowestAllTime;
   double? get lowestLast30Days => _analytics.lowestLast30Days;
   double? get lowestLast7Days => _analytics.lowestLast7Days;
+
+  /// The last year of weekly weight change, oldest first, for the Home heatmap.
+  List<WeeklyWeightChange> get weeklyChanges => _analytics.weeklyChanges();
 }
--- a/lib/ui/pages/home_page.dart
+++ b/lib/ui/pages/home_page.dart
@@ -5,6 +5,7 @@ import 'package:food_locker/features/weight/data/weight_manager.dart';
 import 'package:food_locker/ui/widgets/add_weight_dialog.dart';
 import 'package:food_locker/ui/widgets/app_version_label.dart';
 import 'package:food_locker/ui/widgets/history_range_selector.dart';
+import 'package:food_locker/ui/widgets/weekly_change_heatmap.dart';
 import 'package:food_locker/ui/widgets/weight_history_tile.dart';
 import 'package:provider/provider.dart';
 
@@ -17,6 +18,7 @@ class HomePage extends StatelessWidget {
     final weightManager = context.watch<WeightManager>();
     final history = weightManager.history;
     final range = weightManager.historyRange;
+    final weeklyChanges = weightManager.weeklyChanges;
 
     return Scaffold(
       body: CustomScrollView(
@@ -44,6 +46,13 @@ class HomePage extends StatelessWidget {
               ),
             ),
           ),
+          if (weeklyChanges.any((week) => week.hasData))
+            SliverToBoxAdapter(
+              child: Padding(
+                padding: const EdgeInsets.symmetric(horizontal: 16.0),
+                child: WeeklyChangeHeatmap(weeks: weeklyChanges),
+              ),
+            ),
           SliverToBoxAdapter(
             child: Padding(
               padding: const EdgeInsets.fromLTRB(16.0, 16.0, 16.0, 8.0),
--- a/lib/ui/widgets/weekly_change_color.dart
+++ b/lib/ui/widgets/weekly_change_color.dart
@@ -1,0 +1,31 @@
+import 'package:flutter/material.dart';
+import 'package:food_locker/features/weight/data/weekly_weight_change.dart';
+
+/// The heatmap's colour for a week that reported no change.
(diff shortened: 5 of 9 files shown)


igorbasko01:
**Nothing blocks merge.** The change is genuinely additive — no model, `typeId`, generated file, Drift schema, backup codec, or version file is touched, and the analytics are correct against every decision the issue resolved. I checked the week bucketing, the Sunday offset, the DST-safe walk, the 4-entry threshold, the bucket edges in both units, and the ordering, and found no defect. What follows is four low-severity notes, none of which need fixing before this lands.

### Invariants checked (all clean)

- **Generated code / Hive typeIds / Drift schema** — untouched, and correctly so: the grid is derived at read time from stored `Weight` rows.
- **Day granularity** — `_weekStart` normalises via `DateTime(y, m, d - weekday % 7)`, and week-start keys are built the same way on both sides of the map lookup, so they compare equal even in a zone where local midnight is skipped by a DST shift. The one-entry-per-day repository key also means "≥ 4 entries" is genuinely "≥ 4 distinct days".
- **Mutation shape / derived-not-stored** — read-only path; nothing new is persisted and nothing bypasses the repository.
- **Interfaces over stores** — `WeightAnalytics` reads `WeightRepository`, matching the existing `lowestAllTime` pattern. No new wiring needed in `main.dart`.
- **Backup compatibility** — no new store or CSV column.
- **Versioning** — `pubspec.yaml` and `CHANGELOG.md` untouched; `feat:` is the right prefix for a user-visible feature.
- **Tests** — mirror `lib/`, use `InMemoryWeightRepository`, and cover the behaviour rather than the file. The boundary cases I would have asked for (3 vs 4 entries, the Saturday→Sunday split, out-of-grid and future-dated entries, every bucket edge in kg and lb, exact 0.0) are all there.

### Notes (non-blocking)

**1. The grid size and the week count are two independent constants.** `lib/ui/widgets/weekly_change_heatmap.dart:12-13` fixes 4 × 13, and `lib/features/weight/data/weight_analytics.dart:11` fixes `heatmapWeeks = 52`. `_colorAt` only ever indexes 0…51, and the list is oldest-first — so if `heatmapWeeks` is ever raised to 53, it is the *newest* weeks that fall off the end: the current week silently disappears and the last cell shows a week-old value, with no test failing (`weekly_change_heatmap_test.dart:47` asserts `rows * columns == 52` but nothing ties that to `heatmapWeeks`). Deriving one from the other, or asserting they match, would close it.

**2. `weeklyChanges(weeks: 0)` throws.** `weight_analytics.dart:35` calls `weekStarts.first` on the list `_weekStartsEndingAt` returns, which is empty when `weeks == 0` — `Bad state: No element` rather than an empty list. Unreachable from production (the default is 52) but it is a public parameter on a public method.

**3. A mixed-unit week produces a nonsense delta.** `_change` computes `last.value - first.value` with no conversion, then labels it with the *last* entry's unit (`weight_analytics.dart:58-63`). The add dialog hard-codes `WeightUnit.kilograms`, but `WeightBackupCodec` parses the unit column by name (`weight_backup_codec.dart:67-70`), so restoring an archive with a `pounds` row is a real path to a week holding both units — e.g. 180 lb Monday and 81.6 kg Friday reads as a 98.4 kg loss and paints the darkest green. The existing lowest-weight analytics ignore units the same way, so this is latent rather than a regression, but the heatmap makes it visible.

**4. The heatmap recomputes on every `HomePage` build.** `home_page.dart:21` reads `weightManager.weeklyChanges`, which runs `getAllWeights()` plus a full sort of the year on each rebuild, and `HomePage` watches the manager — so every weigh-in mutation pays for it twice over (once for the `any(...)` visibility check, once for the grid). Fine at today's data sizes; worth remembering alongside the pagination TODO already sitting in `WeightManager`.

### On the PR body's stated decisions

The three judgement calls the description flags — placing the grid under the title block now that #112 removed the streak banners, writing the date math inline instead of reviving `DateTimeCalendar`, and hiding the grid until some week has data — all read as the right call to me, and the last one is covered by a test either way, so it is cheap to reverse if you would rather always show the frame.

---

igorbasko01:
Acted on the review in `978ba24`. Two of the four notes were real and inside this PR's own code; the other two are pre-existing design questions I've deliberately left for you rather than widen the PR.

**Fixed**

- **Grid shape vs. week count (note 1).** Real, and the silent-failure direction the reviewer identified is the bad one: the list is oldest-first, so a grid smaller than `heatmapWeeks` drops the *newest* weeks — the current one included — with nothing failing. The test now asserts `WeeklyChangeHeatmap.rows * columns == WeightAnalytics.heatmapWeeks` instead of a hardcoded `52`, so changing either constant alone breaks the build.
- **`weeklyChanges(weeks: 0)` (note 2).** Now `assert(weeks > 0)`, matching what `DateRange` already does for its own span, plus a test. It replaces an opaque `Bad state: No element`.

**Left alone, on purpose**

- **Mixed-unit weeks (note 3).** Correctly identified as latent rather than a regression. Deciding what a kg-then-lb week *means* — convert, or refuse a delta — is a domain call beyond #105, which explicitly treats the lb column as forward-looking while `AddWeightDialog` hard-codes kilograms. Worth its own issue.
- **Recompute per build (note 4).** Same shape as the pagination `TODO` already sitting on `WeightManager._weights`: caching it needs an invalidation story tied to every mutation path, which is a design change, not a review fix. Also worth its own issue.

Neither of those blocks merge, and neither is a behaviour change this PR introduced.


---

lib/features/weight/data/weight_analytics.dart
```diff
@@ -14,4 +19,62 @@ class WeightAnalytics {
...
+      if (start.isAfter(weekStarts.last)) continue;
+      entriesByWeek.putIfAbsent(start, () => []).add(weight);
+    }
+
+    return [
+      for (final start in weekStarts) _change(start, entriesByWeek[start]),
+    ];
+  }
+
+  static WeeklyWeightChange _change(DateTime weekStart, List<Weight>? entries) {
```
igorbasko01, line 49: why are all of the new private methods are static?

lib/features/weight/data/weight_analytics.dart
```diff
@@ -14,4 +19,62 @@ class WeightAnalytics {
...
+  /// The Sunday opening [date]'s week, at local midnight. Dart counts Mon=1…
+  /// Sun=7, so `% 7` maps Sunday to a zero offset.
+  static DateTime _weekStart(DateTime date) =>
+      DateTime(date.year, date.month, date.day - (date.weekday % 7));
+
+  /// [weeks] consecutive week starts, oldest first, the last being [asOf]'s.
+  ///
+  /// Stepping by calendar components rather than `Duration(days: 7)` keeps the
+  /// walk on Sundays across a daylight-saving shift.
+  static List<DateTime> _weekStartsEndingAt(DateTime asOf, int weeks) {
```
igorbasko01, line 73: does it return the newest week first?

igorbasko01:
please build a test apk after reviewing the comments.

lib/features/weight/data/weight_analytics.dart
igorbasko01: No — **oldest first**, with the current week last.

`back` counts *down*, so the first element is the furthest in the past and the last is `back == 0`, the week `asOf` falls in:

```dart
for (var back = weeks - 1; back >= 0; back--)
  DateTime(current.year, current.month, current.day - 7 * back),
```

That order is the grid's order: `weeklyChanges` maps straight over `weekStarts`, and `WeeklyChangeHeatmap` fills row-major, so oldest lands top-left and the in-progress week lands in the last cell — which is what #105 asked for. Pinned by `returns 52 weeks, oldest first, ending with the current week` and `the current week takes the last cell`.

Fair thing to ask, though, because **the two orderings in this function genuinely differ**: `getWeightsSince` on line 36 returns weigh-ins *newest* first. It doesn't leak into the result — that loop only buckets into a map, and `_change` sorts each week's entries ascending before taking `first`/`last` — but having both directions in one function is a real readability trap. If you'd like, I can rename `_weekStartsEndingAt` to something that says the direction out loud (`_weekStartsOldestFirst`, ending at `asOf`); say the word and I'll push it.


---

lib/features/weight/data/weight_analytics.dart
igorbasko01: Because none of the three touches instance state — they're pure functions of their arguments. `_change`, `_weekStart` and `_weekStartsEndingAt` never read `_weightRepository`; only `weeklyChanges` itself does, on line 36.

`static` is the compiler-enforced version of that claim. It keeps the repository read in one visible place instead of letting a later edit quietly reach for it from inside a helper that reads as pure, and it makes the date math checkable by argument alone.

The contrast is in this class's own history: the `_isOvereaten` / `_calculateCurrentStreak` helpers that #112 removed were instance methods precisely *because* they went through `_weightRepository`:

```dart
bool? _isOvereaten(DateTime date) {
  final weightOnDay = _weightRepository.getWeightForDay(date);
  ...
```

So the split is meant to be informative rather than stylistic — instance when it needs the store, static when it doesn't.

That said, it's a judgment call with no behavioural difference. If you'd rather they match the instance-method shape the class used before, dropping the three `static` keywords is the whole change and I'll push it.


---

igorbasko01:
Comments reviewed and test APK built. 📱

**[⬇️ food_locker-test.apk](https://github.com/igorbasko01/food-locker/releases/download/test-builds/food_locker-test.apk)** — `1.18.0-pr114.10`, built from `978ba24` (this PR's head) via the `Build Test APK` workflow ([run](https://github.com/igorbasko01/food-locker/actions/runs/33361144935)). Installs beside the production app as **FoodLocker Test**, own empty data.

Since the heatmap needs ≥ 4 weigh-ins in a week before a cell colours in, a fresh install shows no grid at all — that's the hidden-when-empty behaviour, not a broken build. To see it with real data, export a backup from the production app's Settings tab and import it here.

**What to look for**

- The grid sits directly under the "Weight Locker" title block, above the history header.
- 4 rows × 13 columns, oldest top-left → newest bottom-right; the **last cell is the week in progress**, so it stays gray until that week reaches 4 weigh-ins.
- Green = lost that week, red = gained, darker = bigger change. A flat week is the faintest green.
- Gray means one thing only: fewer than 4 weigh-ins that week.

**On your two questions** — both answered inline, no code change needed for either:

- `_weekStartsEndingAt` returns **oldest first**, current week last (`back` counts down). Offered a rename to say the direction out loud, since `getWeightsSince` on the line above returns the opposite order.
- The three new private helpers are `static` because none reads `_weightRepository` — the same reason the old `_isOvereaten` *wasn't*. Happy to drop the keyword if you'd rather match the previous shape.

Both are one-word/one-line changes waiting on your call; say which way and I'll push and rebuild.


---

Issue #116, opened 2026-08-31
feat: show the weekday alongside every displayed date


igorbasko01:
## Problem

Dates are shown as bare numbers throughout the app (`7/14`, `2026-03-08`), so it takes mental arithmetic to tell whether a weigh-in or a big bite day was a weekday or a weekend. The weekday is often the most useful part of the date when reading these trends.

## Depends on #117

#117 consolidates the three hand-rolled `YYYY-MM-DD` sites onto `lib/core/date_format.dart`. With that done, this issue is a change to the shared helpers plus their call sites, rather than a hunt through the widgets.

## Proposal

Add the short weekday name — `DateFormat.E`, three letters (`Sun`, `Mon`, `Sat`) — to the dates, keeping the existing locale-aware numeric ordering.

Extend `lib/core/date_format.dart` with weekday-carrying variants alongside `shortDate`/`fullDate`, then use them at the call sites.

**Charts get the weekday in tooltips only — never on axis labels.** The axes stay numeric-only (`shortDate`); the bites axis alone draws ~30 labels in a row.

Call sites:

- `lib/ui/widgets/weight_history_tile.dart` and `lib/ui/pages/weight_page.dart` — the weight rows
- `lib/ui/widgets/add_weight_dialog.dart` — the picked date
- `lib/ui/widgets/weight_chart.dart` — tooltip only
- `lib/ui/widgets/daily_bites_chart.dart` — tooltip and accessibility summary only
- `lib/ui/pages/bite_analytics_page.dart` — the `30-day max` stat tile sub-label and the meal-breakdown card title for a non-today day
- the heatmap tooltip from #115, once that lands

`DateFormat.EEEEE` (narrow) is **not** an option, in case it looks tempting for the tighter spots: English narrow names are `S M T W T F S`, so Sunday and Saturday are both `S` and Tuesday and Thursday both `T` — and weekend vs weekday is the whole point of this issue.

## Notes

- Keep everything going through `date_format.dart` rather than formatting inline, so locale handling stays in one place (`initializeDateFormatting()` is already called in `main.dart`).
- The `30-day max` stat tile sub-label is the tightest of the call sites — check it visually before committing to the weekday there.
- Update the affected widget tests, which assert on numeric-only date strings.

PR #127, opened 2026-08-31
refactor: route every displayed date through core/date_format.dart

### LINKED ISSUE #117: refactor: route every displayed date through core/date_format.dart
## Problem

Date rendering is split in two. `lib/core/date_format.dart` holds the locale-aware helpers (`shortDate`, `fullDate`), used by the charts and the bite analytics page — but three UI sites hand-roll an ISO-ish `YYYY-MM-DD` string instead, ignoring the device locale:

- `lib/ui/widgets/weight_history_tile.dart:18`
- `lib/ui/pages/weight_page.dart:92`
- `lib/ui/widgets/add_weight_dialog.dart:74`

So the same date reads as `3/8/2026` on a chart tooltip and `2026-03-08` in the weight row right under it.

## Proposal

Move all three onto the shared locale-aware helpers, so every date in the app is formatted in one place and one style.

The rule: **`shortDate` where space is tight, `fullDate` everywhere else.**

- `shortDate` — chart axis labels (`weight_chart.dart:76`, `daily_bites_chart.dart:247`) and the `30-day max` stat tile sub-label (`bite_analytics_page.dart:200`).
- `fullDate` — everything else: the three sites above, tooltips, and the meal-breakdown card title (`bite_analytics_page.dart:287`, currently `shortDate` but has the room).

## Notes

- `initializeDateFormatting()` is already called in `main.dart`; the helpers take an optional `locale` override so tests stay deterministic.
- Groundwork for #116 (adding the weekday to every date) — with formatting consolidated, that becomes a change to `date_format.dart` plus its call sites instead of a hunt through the widgets.
- Existing widget tests assert on the `YYYY-MM-DD` strings and will need updating.

igorbasko01:
## What changed

Every date **and time** the app displays now goes through `lib/core/date_format.dart`, so the same day reads identically in a weight row and in the chart tooltip above it, and clock times follow the device locale.

The three hand-rolled `YYYY-MM-DD` sites named in the issue now call `fullDate`:

- `lib/ui/widgets/weight_history_tile.dart:19`
- `lib/ui/pages/weight_page.dart:91` (the stale `// Format date nicely` comment went with it)
- `lib/ui/widgets/add_weight_dialog.dart:75`

And the one site the issue asked to re-bucket:

- `lib/ui/pages/bite_analytics_page.dart:287` — the meal-breakdown card title moves `shortDate` → `fullDate`, since it has the room.

Left on `shortDate` per the rule (space is tight): the two chart axis labels (`weight_chart.dart:76`, `daily_bites_chart.dart:247`) and the `30-day max` stat tile sub-label. Tooltips and the bites accessibility summary were already on `fullDate` and are untouched.

The `date_format.dart` header and the `fullDate` doc comment said "for charts and stat tiles" / "for chart tooltips", which stopped being true with this change — both now state the app-wide rule instead.

### Follow-up commit — the two review items you asked for

- **Time locale (`meal_breakdown_list.dart:160-166`).** That file hand-rolled `h:mm AM/PM` — the last display formatter in `lib/` outside `date_format.dart`, and a real bug for anyone on a 24-hour locale, who still got `1:30 PM`. A new `shortTime` helper (`DateFormat.jm`) joins `shortDate`/`fullDate`, and the meal rows use it: `1:30 PM` under `en_US`, `13:30` under `en_GB`.
- **The date pass-through (`bite_analytics_page.dart:200`).** `_formatDay` was a single-caller alias for `shortDate`; the `30-day max` sub-label now calls the helper directly and the alias is gone.

`lib/` now contains no date or time formatting outside `core/date_format.dart`.

## How the issue is met

- **All three hand-rolled sites moved onto the shared helpers.** Done, as listed above — no `padLeft`-built date strings remain in `lib/`.
- **`shortDate` where space is tight, `fullDate` everywhere else.** Applied exactly as the issue specifies, including the meal-breakdown title's move to `fullDate`.
- **Existing widget tests assert on the `YYYY-MM-DD` strings and will need updating.** `test/ui/pages/weight_page_test.dart` and `test/ui/pages/home_page_test.dart` each carried a private `_dateLabel` helper that rebuilt the ISO string; both are deleted and the assertions now call `fullDate` directly, so the tests read the date through the same helper the widget does.

## Tests

- `test/ui/pages/weight_page_test.dart`, `test/ui/pages/home_page_test.dart` — updated to expect the locale-aware string.
- `test/ui/pages/bite_analytics_page_test.dart` — the day-selection test now asserts the non-today meal card is titled `fullDate(yesterday)`, which is the behaviour this PR changes; previously it only checked that the `Today's meals` title was gone.
- `test/ui/widgets/add_weight_dialog_test.dart` — new, covering the dialog's date button. It was the one changed site with no test at all.
- `test/core/date_format_test.dart` — `shortTime` is pinned against both conventions: 12-hour with a day period under `en_US`, `13:30` under `en_GB`.
- `test/ui/widgets/meal_breakdown_list_test.dart` — the two meal-span assertions were `'8:00 AM – 8:11 AM'` literals. They now build the expected span through `shortTime`, because current CLDR puts a narrow no-break space (U+202F) before `AM`/`PM`, so a plain-space literal no longer matches what `intl` renders.

## Decisions

- **Tests compute the expected string with `fullDate(date)` / `shortTime(at)` rather than pinned-locale literals.** Both the widget and the test then resolve the locale the same way, so the assertions hold on any runner. `test/core/date_format_test.dart` covers the format contracts themselves with explicit `en_US`/`en_GB` overrides.
- **No weekday added anywhere.** That is #116, which explicitly builds on this consolidation; this PR is the groundwork only.
- **Title still says `refactor:`, matching #117's own title** — flagged because the squashed subject is the only lever and `refactor:` is hidden from the release-please changelog, while this PR does change what users see (dates, and now times on 24-hour locales). Worth overriding to `fix:` at merge if you want it in the notes.

## Verification

Flutter is not installed in this session and `CLAUDE.md` says not to install it, so `flutter analyze` and `flutter test` were **not** run locally — verification is deferred to the `flutter_ci.yml` checks on this PR. `analyze_and_test` is green on both commits.

Closes #117

### DIFF
--- a/lib/core/date_format.dart
+++ b/lib/core/date_format.dart
@@ -2,10 +2,11 @@ import 'dart:ui';
 
 import 'package:intl/intl.dart';
 
-/// Locale-aware numeric date formatting for every date the app displays.
+/// Locale-aware date and time formatting for everything the app displays.
 ///
 /// Dates are drawn in the device locale's field order — `7/14` under `en_US`,
-/// `14/7` under `en_GB` — as numbers, never month names. Both helpers default to
+/// `14/7` under `en_GB` — as numbers, never month names, and times follow the
+/// locale's clock convention. Every helper defaults to
 /// [PlatformDispatcher.instance.locale]; an explicit [locale] override keeps them
 /// deterministic under test. Non-`en_US` locales need `initializeDateFormatting()`
 /// (called once at startup in `main.dart`).
@@ -17,5 +18,9 @@ String shortDate(DateTime date, [String? locale]) =>
 /// Numeric year/month/day in [locale] order, the default where space allows.
 String fullDate(DateTime date, [String? locale]) =>
     DateFormat.yMd(locale ?? _deviceLocale).format(date);
+
+/// The clock time, 12- or 24-hour as [locale] dictates.
+String shortTime(DateTime at, [String? locale]) =>
+    DateFormat.jm(locale ?? _deviceLocale).format(at);
 
 String get _deviceLocale => PlatformDispatcher.instance.locale.toString();
--- a/lib/core/date_format.dart
+++ b/lib/core/date_format.dart
@@ -2,7 +2,7 @@ import 'dart:ui';
 
 import 'package:intl/intl.dart';
 
-/// Locale-aware numeric date formatting for charts and stat tiles.
+/// Locale-aware numeric date formatting for every date the app displays.
 ///
 /// Dates are drawn in the device locale's field order — `7/14` under `en_US`,
 /// `14/7` under `en_GB` — as numbers, never month names. Both helpers default to
@@ -14,7 +14,7 @@ import 'package:intl/intl.dart';
 String shortDate(DateTime date, [String? locale]) =>
     DateFormat.Md(locale ?? _deviceLocale).format(date);
 
-/// Numeric year/month/day in [locale] order, for chart tooltips.
+/// Numeric year/month/day in [locale] order, the default where space allows.
 String fullDate(DateTime date, [String? locale]) =>
     DateFormat.yMd(locale ?? _deviceLocale).format(date);
 
--- a/lib/ui/pages/bite_analytics_page.dart
+++ b/lib/ui/pages/bite_analytics_page.dart
@@ -284,7 +284,7 @@ class _MealBreakdownCard extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
-    final title = isToday ? 'Today\'s meals' : shortDate(breakdown.day);
+    final title = isToday ? 'Today\'s meals' : fullDate(breakdown.day);
     return Card(
       elevation: 0,
       margin: const EdgeInsets.fromLTRB(16.0, 0, 16.0, 16.0),
--- a/lib/ui/pages/bite_analytics_page.dart
+++ b/lib/ui/pages/bite_analytics_page.dart
@@ -182,7 +182,7 @@ class _StatTilesRow extends StatelessWidget {
               child: StatTile(
                 label: '30-day max',
                 value: max == null ? '—' : max.count.toString(),
-                subLabel: max == null ? null : _formatDay(max.day),
+                subLabel: max == null ? null : shortDate(max.day),
               ),
             ),
           ],
@@ -194,10 +194,6 @@ class _StatTilesRow extends StatelessWidget {
   /// A whole-bite figure, with `—` for the no-qualifying-day case (average 0).
   static String _formatAverage(double average) =>
       average == 0 ? '—' : average.toStringAsFixed(0);
-
-  /// A day in the device locale's numeric order, matching the daily-bites
-  /// chart's axis labels.
-  static String _formatDay(DateTime day) => shortDate(day);
 }
 
 /// A meals summary: today's meal count, the 30-day average meals per day, and
--- a/lib/ui/pages/weight_page.dart
+++ b/lib/ui/pages/weight_page.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:food_locker/core/date_format.dart';
 import 'package:food_locker/features/weight/data/weight.dart';
 import 'package:food_locker/features/weight/data/weight_manager.dart';
 import 'package:food_locker/ui/widgets/add_weight_dialog.dart';
@@ -88,8 +89,7 @@ class WeightPage extends StatelessWidget {
                   delegate: SliverChildBuilderDelegate(
                     (context, index) {
                       final item = history[index];
-                      // Format date nicely
-                      final dateStr = '${item.date.year}-${item.date.month.toString().padLeft(2, '0')}-${item.date.day.toString().padLeft(2, '0')}';
+                      final dateStr = fullDate(item.date);
 
                       return Dismissible(
                         key: ValueKey(item.date),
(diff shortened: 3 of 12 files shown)


igorbasko01:
**Verdict: nothing blocks merge.** The change does exactly what #117 asks, the three hand-rolled `YYYY-MM-DD` sites are gone, and no repo invariant is touched. One advisory note on the commit prefix, and a few small observations that are fine to ignore.

## Invariants — all clear

- **Generated code / Hive typeIds / Drift schema** — untouched. No `@HiveType`, `@HiveField`, `@DriftDatabase` table, `*.g.dart` or `hive_registrar.g.dart` in the diff, so nothing is stale and nothing needs a `schemaVersion` bump.
- **Day granularity** — the change is display-only. `weight_page.dart:95` still keys the `Dismissible` on `ValueKey(item.date)` and `deleteWeight(item.date)` still receives the raw `DateTime`; no formatted string feeds a key, a comparison, or a repository call. `add_weight_dialog.dart` still pops the raw `_selectedDate`, so the one-entry-per-day normalization downstream is unaffected.
- **Mutation shape** — no mutation paths in the diff.
- **Backup compatibility** — no codec touched, and correctly so: `serialization_service.dart:25-29` keeps its own `padLeft` timestamp for the archive *filename*, which must stay locale-independent. Routing that through `fullDate` would have put `/` in a filename and broken naming for existing users — good call leaving it alone.
- **Versioning** — no manual `pubspec.yaml` or `CHANGELOG.md` edit.
- **Tests** — mirror `lib/`, and `test/ui/widgets/add_weight_dialog_test.dart` lands in the right place. `hive_ce_migration_test.dart` untouched.

I also checked for assertions this would strand: no test outside the four updated files asserts an ISO date string (`grep` for `padLeft` in `test/` comes back empty), and there is no `weight_history_tile_test.dart` to update — that widget's date is covered through `home_page_test.dart`. That matches CI passing.

## Findings

### 1. `refactor:` hides a user-visible change from the changelog — advisory

The squashed subject computes the next version. `release-please-action` here runs on defaults (`release-type: dart`, no config file), where `refactor` is a hidden type: it will not produce a `CHANGELOG.md` entry and will not earn a minor.

But this is not an internal-only change. Concretely: a user on `en_US` who opens the Weight tab today sees `2026-03-08` in every history row and on the add-weight dialog's date button; after this ships they see `3/8/2026`. Every date in the weight feature changes shape for every existing user, and the release notes for that version will say nothing about it.

#117 frames the problem as a defect ("the same date reads as `3/8/2026` on a chart tooltip and `2026-03-08` in the weight row right under it"), which reads as `fix:`. Worth deciding deliberately, since the merge-time subject is editable and the prefix is the only lever. Not a code change either way.

### 2. The updated widget assertions can't catch a `fullDate` regression

`weight_page_test.dart:72`, `home_page_test.dart:108`, `bite_analytics_page_test.dart:326` and the new dialog test all compute the expected string by calling `fullDate` — the same function the widget calls. If `fullDate` regressed to emitting ISO strings tomorrow, all four would still pass.

That is the right trade here and the PR body already reasons it out: they assert *that the widget routes through the shared helper*, which is precisely what this PR is about, and `test/core/date_format_test.dart` pins the actual format contract with explicit `en_US`/`en_GB` overrides. Flagging only so the division of labour is on the record — the format contract lives in exactly one test file now, so that file is the one to guard.

### 3. `add_weight_dialog_test.dart` covers the initial render only — minor

The test asserts the button's label for the `initialDate`, which does cover the changed line. It does not exercise the `setState` path after `showDatePicker` returns a new date (`add_weight_dialog.dart:68-76`). Tapping through a `showDatePicker` in a widget test is fiddly enough that the cost/benefit is defensible for a formatting change; noting it as the one uncovered branch of the file this PR added a test for.

## Not in scope, worth knowing

- `lib/ui/widgets/meal_breakdown_list.dart:160-166` — `_formatTime` still hand-rolls `h:mm AM/PM`, so a user on a 24-hour locale sees `8:05 PM` in the meal breakdown next to a now-locale-aware date. #117 is explicitly about dates, so this is correctly out of scope, but it is the last display formatter in `lib/` bypassing `intl`, and it sits right under a date this PR just made locale-aware. Reasonable to fold into #116 or a follow-up.
- `lib/ui/pages/bite_analytics_page.dart:200` — `static String _formatDay(DateTime day) => shortDate(day);` is now a pure pass-through alias with a single caller. Pre-existing and trivial; inlining it would remove one hop. Keeping it as-is is a fine call too.

---

igorbasko01:
Handle the time locale in 
lib/ui/widgets/meal_breakdown_list.dart:160-166

And the date pass through comment.

igorbasko01:
Both done in `e04ec52`.

**Time locale** — `meal_breakdown_list.dart` was building `h:mm AM/PM` by hand, so a 24-hour locale still read `1:30 PM`. Added `shortTime` (`DateFormat.jm`) next to `shortDate`/`fullDate` and pointed the meal rows at it: `1:30 PM` under `en_US`, `13:30` under `en_GB`. That was the last display formatter in `lib/` outside `core/date_format.dart`.

**The date pass-through** — `_formatDay` was a single-caller alias for `shortDate`; the `30-day max` sub-label calls `shortDate` directly now and the alias is gone.

One thing worth knowing about the tests: the two meal-span assertions were `'8:00 AM – 8:11 AM'` literals, and they had to change. Current CLDR puts a narrow no-break space (U+202F) before `AM`/`PM`, so `intl` no longer renders a plain space there and the old literals would have gone red. They build the expected span through `shortTime` now, with `date_format_test.dart` pinning the actual `en_US`/`en_GB` contract.

---

PR #131, opened 2026-09-02
feat: explain the Home heatmap cells and gate weeks on span

### LINKED ISSUE #115: feat: explain the Home weekly-change heatmap cells — tooltip, and a span-based no-data gate
## Problem

The 52-week heatmap on Home (`lib/ui/widgets/weekly_change_heatmap.dart`) is currently unlabelled and non-interactive by design: cells are drawn oldest top-left to newest bottom-right, coloured green for a losing week and red for a gaining week, with the shade coming from the magnitude bucket in `WeeklyWeightChange.level`.

As a user I can't tell **why** a square is red, or **what was compared** to arrive at that colour.

Worse, a lot of squares are grey that shouldn't be — see the second part below.

## Part 1 — Tooltip

Each cell shows its week **while the finger is held on it** — the tooltip appears on press and disappears on release. No tap-to-open popup that has to be dismissed.

The tooltip should show:

- **the period the cell covers** — the week start (Sunday, `WeeklyWeightChange.weekStart`) through its Saturday, e.g. `Sun Mar 8 – Sat Mar 14`
- **the two weigh-ins that were compared** — the week's first and last weigh-in, each with its date and value, e.g. `Mon Mar 9: 82.4 kg → Fri Mar 13: 83.0 kg`. This is the part that makes the colour make sense: the delta is **intra-week** (last weigh-in of the week minus the first, per `WeeklyWeightChange`), nothing carries across a week boundary and no two weeks are compared against each other.
- **the resulting change**, signed and with its unit — `+0.6 kg` / `-1.2 kg`
- **a clear no-data state** for weeks that don't clear the gate and for empty cells past the end of the list — these are the grey squares and currently read as "nothing happened", which is misleading

No colour legend needed — showing the delta and the two weigh-ins behind it is enough to explain red vs green.

## Part 2 — Replace the entry-count gate with a span gate

`WeeklyWeightChange.minEntries` is currently 4: a week with fewer than four weigh-ins reports no delta and draws grey.

What that threshold is *actually* trying to guarantee is that the delta **spans** the week rather than measuring a two-day blip. A count is a lossy proxy for that, and in one direction it's plainly wrong:

- **Sunday and Saturday only** — 2 entries, 6-day span. A textbook weekly measurement, discarded as no-data.

(The opposite failure — many entries clustered into one or two days — is *not* possible here: `PersistentWeightRepository.saveWeight` keys on `getKey(date)` = `'y-m-d'`, so the store holds at most one entry per day. Four entries in a Sun–Sat week therefore already implies a span of at least three days. The count gate is under-inclusive only.)

**Proposal:** replace `minEntries` with a minimum **span** — the week's first and last weigh-in at least 3 days apart. This encodes the intent directly, admits genuinely informative sparse weeks, and needs no separate "at least 2 entries" rule since a non-zero span requires two distinct days.

This also fixes a live bug at the bottom end. `WeightAnalytics._change` computes `last.value - first.value` after sorting; with a **single** weigh-in `first == last`, so `delta` is `0.0` rather than null. `level` then returns 1 and `isGain` is false, and the cell renders as **faintest green** — "you lost a tiny bit this week", from one data point. Today `minEntries = 4` masks this; any gate that admits low-entry weeks must handle it explicitly, and a span gate does (a single entry has span 0).

## Notes

- `WeeklyWeightChange` today carries only `weekStart`, `delta` and `unit`, so surfacing the compared weigh-ins means adding the endpoints (dates + values) to the model and populating them where the weeks are built (`WeightAnalytics` / `WeightManager.weeklyChanges`). Keep it a plain data addition — the delta stays computed as it is now.
- Dates should go through `lib/core/date_format.dart` so they stay locale-aware, and should include the weekday (see #116).
- Add a `Semantics` label per cell as well, so the information is reachable without hover/long-press.
- `minEntries` is referenced only in `WeightAnalytics._change` and the `WeeklyWeightChange` doc comments — no test asserts on it, so swapping the gate is contained.
- #119 reuses this gate for its weekly-change stat tile; whatever lands here is the shared rule.
- Tests: extend `test/ui/widgets/weekly_change_heatmap_test.dart` to cover the tooltip text for a gaining week, a losing week, and a no-data week, plus model/analytics tests for the endpoint fields and the new gate — specifically a Sun+Sat two-entry week (now passes), a Mon+Tue two-entry week (still grey), and a single-entry week (grey, not faint green).

igorbasko01:
## What changed

**Part 1 — reading a cell.** The grid handles its own touch and draws its own readout, the way the charts do. Press a cell and it names the week; drag and the readout follows your finger from cell to cell; lift, or leave the grid, and it ends.

```
Sun, 3/8 – Sat, 3/14
Tue, 3/10: 82.4 kg → Sat, 3/14: 83.0 kg
+0.6 kg
```

- the period the cell covers — the week's Sunday through its Saturday;
- the two weigh-ins that were compared, each with date and value, which is what makes the colour make sense;
- the resulting change, signed and with its unit;
- for a week with no delta, `No two weigh-ins 3 days apart` — true whether the week held nothing at all or two weigh-ins a day apart — and for a cell past the end of the list, `No data`. Grey no longer reads as "nothing happened".

The readout opens **above** the held cell (offset from the cell's centre, so it clears the fingertip), dropping below only when a cell sits too near the top of the screen to fit. Every cell also carries the same lines as a `Semantics` label, so the information is reachable without holding anything. Dates go through `lib/core/date_format.dart`, so they stay locale-aware and carry the weekday.

**Part 2 — the span gate.** `WeeklyWeightChange.minEntries = 4` is replaced by `minSpanDays = 3`: a week reports a delta only when its first and last weigh-in lie at least three days apart. A Sunday-and-Saturday pair (2 entries, 6-day span) now reports its change instead of drawing grey; a Monday-and-Tuesday pair still doesn't.

This also fixes the bug the old gate was masking: a single weigh-in made `last.value - first.value == 0.0`, which gave `level == 1` and `isGain == false` — the cell rendered as the faintest green, "you lost a tiny bit", from one data point. A lone weigh-in spans zero days, so it is now correctly no-data.

**Model.** `WeeklyWeightChange` takes the week's weigh-ins and derives everything from them:

```dart
factory WeeklyWeightChange({required DateTime weekStart, List<Weight> entries = const []})

double? get delta => hasData ? last!.value - first!.value : null;
bool get hasData => _spanInDays >= minSpanDays;
```

`weekStart` is the week's identity — an empty week still has a cell to fill and a period to name — and a constructor assert holds the entries to that week. `WeightAnalytics` is left with the job only it can do: reading the repository and working out which week each weigh-in belongs to. No Hive model, no `typeId`, no Drift table, no backup-codec impact.

## How each acceptance criterion is met

| From the issue | Where |
|---|---|
| Readout while the cell is held, gone on release | `_WeeklyChangeHeatmapState`; tests *holding a cell names its week and the weigh-ins behind it*, *dragging off the grid ends the readout* |
| Period the cell covers | `weeklyChangeSummary` first line; test *a gaining week names its period, its weigh-ins and its change* |
| The two weigh-ins compared, with dates and values | `first`/`last` on the model + summary second line; tests *carries the two weigh-ins it compared*, *the weigh-ins carry the week's own unit* |
| Resulting change, signed, with unit | summary last line; tests for `+0.6 kg`, `-1.2 kg`, `0.0 kg` |
| Clear no-data state, gated weeks and empty cells alike | tests *a week without a delta says so instead of reading as flat*, *a cell past the end of the grid has no week to name* |
| `minEntries` → minimum span | `WeeklyWeightChange.minSpanDays` |
| Sun+Sat two-entry week passes | test *a Sunday and a Saturday alone span the whole week* |
| Mon+Tue two-entry week still grey | test *needs the week's weigh-ins three days apart* |
| Single-entry week grey, not faint green | test *a lone weigh-in reports nothing rather than a flat week* |
| Dates via `core/date_format.dart`, with weekday | `weeklyChangeSummary` uses `shortDateWithWeekday` |
| `Semantics` label per cell | test *each cell is readable without holding it* |

Beyond the issue, from review: dragging moves the readout between cells (test *dragging carries the readout to the next cell*).

- **The grid owns the gesture, not the cells.** One `Listener` over the whole heatmap, with the cell under the finger computed from `LayoutBuilder`'s cell size rather than widget hit-testing. This is what makes drag-between-cells possible, and it removes the 52 `Tooltip`s and `GlobalKey`s an earlier revision needed.
- **The readout is recomputed from the finger's position on every move**, so a drag that also scrolls the page keeps it glued to the cell under the finger instead of stranding it — which is why there's no touch-slop dismissal any more.
- **Span counted on UTC midnights**, so a daylight-saving shift inside the week can't turn a 3-day span into 2 days 23 hours.
- **One no-data line for both cases.** A week with nothing logged and a week whose weigh-ins sat too close say the same thing, because the model distinguishes them only by absence; telling them apart would mean carrying an entry count, which is more than this issue asks for.
- **`Weight` keeps its identity equality.** The model's `==` compares weigh-ins on date, value and unit rather than changing how `Weight` itself compares — that's a Hive model used well outside this feature.
- **No colour legend**, per the issue — the delta and the two weigh-ins behind it explain red vs green.

## Verification

Flutter is not installed in this session and the repo's `CLAUDE.md` says not to install it, so `flutter analyze` and `flutter test` were **not** run locally — verification is deferred to the `flutter_ci.yml` checks on this PR. The change was also reviewed by a separate agent working only from the diff, and tried on a phone via the `test-builds` APK, which is where the tooltip-under-the-finger and drag behaviour were caught.

Closes #115

https://claude.ai/code/session_018mtHR2XTR23mGAGsaNQZk5

### DIFF
--- a/lib/features/weight/data/weekly_weight_change.dart
+++ b/lib/features/weight/data/weekly_weight_change.dart
@@ -13,6 +13,11 @@ class WeeklyWeightChange {
     required DateTime weekStart,
     List<Weight> entries = const [],
   }) {
+    assert(
+      entries.every((entry) => _fallsInWeek(entry, weekStart)),
+      'every weigh-in must fall in the week weekStart opens',
+    );
+
     if (entries.isEmpty) return WeeklyWeightChange._(weekStart: weekStart);
 
     var first = entries.first;
@@ -40,6 +45,9 @@ class WeeklyWeightChange {
   static const int minSpanDays = 3;
 
   /// The Sunday opening the week, at local midnight.
+  ///
+  /// The week's own identity, not something [entries] could supply: a week
+  /// nothing was logged in still has a cell to fill and a period to name.
   final DateTime weekStart;
 
   /// The week's earliest and latest weigh-in, null only for a week that holds
@@ -100,6 +108,17 @@ class WeeklyWeightChange {
   /// `Weight` compares by identity, so weigh-ins are matched on what they say.
   static bool _sameWeighIn(Weight? a, Weight? b) =>
       a?.date == b?.date && a?.value == b?.value && a?.unit == b?.unit;
+
+  /// Whether [entry] falls in the seven days [weekStart] opens.
+  static bool _fallsInWeek(Weight entry, DateTime weekStart) {
+    final day = DateTime(entry.date.year, entry.date.month, entry.date.day);
+    final weekAfter = DateTime(
+      weekStart.year,
+      weekStart.month,
+      weekStart.day + 7,
+    );
+    return !day.isBefore(weekStart) && day.isBefore(weekAfter);
+  }
 
   @override
   String toString() =>
--- a/lib/features/weight/data/weekly_weight_change.dart
+++ b/lib/features/weight/data/weekly_weight_change.dart
@@ -4,21 +4,44 @@ import 'package:food_locker/features/weight/data/weight.dart';
 ///
 /// Intra-week only: [delta] is the week's last weigh-in minus its first, so
 /// nothing carries across a week boundary and every week stands on its own. A
-/// week under [minEntries] weigh-ins reports no [delta] at all.
+/// week whose weigh-ins span fewer than [minSpanDays] days reports no [delta]
+/// at all.
 class WeeklyWeightChange {
-  const WeeklyWeightChange({required this.weekStart, this.delta, this.unit});
+  const WeeklyWeightChange({
+    required this.weekStart,
+    this.delta,
+    this.unit,
+    this.firstDate,
+    this.firstValue,
+    this.lastDate,
+    this.lastValue,
+  });
 
-  /// Weigh-ins a week needs before it reports a [delta].
-  static const int minEntries = 4;
+  /// Days a week's first and last weigh-in must lie apart before it reports a
+  /// [delta].
+  ///
+  /// The span states directly what a delta has to measure — the week, not a
+  /// two-day blip: a Sunday and a Saturday qualify on two weigh-ins, a Monday
+  /// and Tuesday never do. A lone weigh-in spans zero days, so its
+  /// `last - first == 0` cannot render as a faint loss.
+  static const int minSpanDays = 3;
 
   /// The Sunday opening the week, at local midnight.
   final DateTime weekStart;
 
-  /// Last weigh-in of the week minus its first, or null under [minEntries].
+  /// Last weigh-in of the week minus its first, or null under [minSpanDays].
   final double? delta;
 
   /// The unit [delta] is expressed in; null exactly when [delta] is.
   final WeightUnit? unit;
+
+  /// The week's first weigh-in, [delta]'s baseline; null when [delta] is.
+  final DateTime? firstDate;
+  final double? firstValue;
+
+  /// The week's last weigh-in, [delta]'s far end; null when [delta] is.
+  final DateTime? lastDate;
+  final double? lastValue;
 
   bool get hasData => delta != null;
 
@@ -42,12 +65,25 @@ class WeeklyWeightChange {
       other is WeeklyWeightChange &&
       other.weekStart == weekStart &&
       other.delta == delta &&
-      other.unit == unit;
+      other.unit == unit &&
+      other.firstDate == firstDate &&
+      other.firstValue == firstValue &&
+      other.lastDate == lastDate &&
+      other.lastValue == lastValue;
 
   @override
-  int get hashCode => Object.hash(weekStart, delta, unit);
+  int get hashCode => Object.hash(
+    weekStart,
+    delta,
+    unit,
+    firstDate,
+    firstValue,
+    lastDate,
+    lastValue,
+  );
 
   @override
   String toString() =>
-      'WeeklyWeightChange(weekStart: $weekStart, delta: $delta, unit: $unit)';
+      'WeeklyWeightChange(weekStart: $weekStart, delta: $delta, unit: $unit, '
+      'first: $firstValue on $firstDate, last: $lastValue on $lastDate)';
 }
--- a/lib/features/weight/data/weekly_weight_change.dart
+++ b/lib/features/weight/data/weekly_weight_change.dart
@@ -15,7 +15,21 @@ class WeeklyWeightChange {
     this.firstValue,
     this.lastDate,
     this.lastValue,
-  });
+  }) : assert(
+         delta == null
+             ? unit == null &&
+                   firstDate == null &&
+                   firstValue == null &&
+                   lastDate == null &&
+                   lastValue == null
+             : unit != null &&
+                   firstDate != null &&
+                   firstValue != null &&
+                   lastDate != null &&
+                   lastValue != null,
+         'a week reports its delta together with the two weigh-ins and the '
+         'unit behind it, or reports nothing at all',
+       );
 
   /// Days a week's first and last weigh-in must lie apart before it reports a
   /// [delta].
--- a/lib/features/weight/data/weekly_weight_change.dart
+++ b/lib/features/weight/data/weekly_weight_change.dart
@@ -2,34 +2,33 @@ import 'package:food_locker/features/weight/data/weight.dart';
 
 /// One calendar week's weight change, as the Home heatmap reads it.
 ///
-/// Intra-week only: [delta] is the week's last weigh-in minus its first, so
-/// nothing carries across a week boundary and every week stands on its own. A
-/// week whose weigh-ins span fewer than [minSpanDays] days reports no [delta]
-/// at all.
+/// Built from the week's own weigh-ins and derives everything from them:
+/// intra-week only, so nothing carries across a week boundary and every week
+/// stands on its own. A week whose weigh-ins span fewer than [minSpanDays]
+/// days reports no [delta] at all.
 class WeeklyWeightChange {
-  const WeeklyWeightChange({
-    required this.weekStart,
-    this.delta,
-    this.unit,
-    this.firstDate,
-    this.firstValue,
-    this.lastDate,
-    this.lastValue,
-  }) : assert(
-         delta == null
-             ? unit == null &&
-                   firstDate == null &&
-                   firstValue == null &&
-                   lastDate == null &&
-                   lastValue == null
-             : unit != null &&
-                   firstDate != null &&
-                   firstValue != null &&
-                   lastDate != null &&
-                   lastValue != null,
-         'a week reports its delta together with the two weigh-ins and the '
-         'unit behind it, or reports nothing at all',
-       );
+  /// The week [weekStart] opens, measured across [entries] — the week's
+  /// weigh-ins, in any order. An empty list is a week nothing was logged in.
+  factory WeeklyWeightChange({
+    required DateTime weekStart,
+    List<Weight> entries = const [],
+  }) {
+    if (entries.isEmpty) return WeeklyWeightChange._(weekStart: weekStart);
+
+    var first = entries.first;
+    var last = entries.first;
+    for (final entry in entries) {
+      if (entry.date.isBefore(first.date)) first = entry;
+      if (entry.date.isAfter(last.date)) last = entry;
+    }
+    return WeeklyWeightChange._(
+      weekStart: weekStart,
+      first: first,
+      last: last,
+    );
+  }
+
+  const WeeklyWeightChange._({required this.weekStart, this.first, this.last});
 
   /// Days a week's first and last weigh-in must lie apart before it reports a
   /// [delta].
@@ -43,21 +42,18 @@ class WeeklyWeightChange {
   /// The Sunday opening the week, at local midnight.
   final DateTime weekStart;
 
+  /// The week's earliest and latest weigh-in, null only for a week that holds
+  /// none. They are the same weigh-in when the week holds exactly one.
+  final Weight? first;
+  final Weight? last;
+
   /// Last weigh-in of the week minus its first, or null under [minSpanDays].
-  final double? delta;
+  double? get delta => hasData ? last!.value - first!.value : null;
 
   /// The unit [delta] is expressed in; null exactly when [delta] is.
-  final WeightUnit? unit;
+  WeightUnit? get unit => hasData ? last!.unit : null;
 
-  /// The week's first weigh-in, [delta]'s baseline; null when [delta] is.
-  final DateTime? firstDate;
-  final double? firstValue;
-
-  /// The week's last weigh-in, [delta]'s far end; null when [delta] is.
-  final DateTime? lastDate;
-  final double? lastValue;
-
-  bool get hasData => delta != null;
+  bool get hasData => _spanInDays >= minSpanDays;
 
   /// Whether the week ended heavier. A flat week counts as not gained.
   bool get isGain => (delta ?? 0) > 0;
@@ -74,30 +70,40 @@ class WeeklyWeightChange {
     return 4;
   }
 
+  /// Calendar days from the first weigh-in to the last, counted on UTC
+  /// midnights so a daylight-saving shift inside the week cannot shorten it.
+  int get _spanInDays {
+    final from = first?.date;
+    final to = last?.date;
+    if (from == null || to == null) return 0;
+    return DateTime.utc(to.year, to.month, to.day)
+        .difference(DateTime.utc(from.year, from.month, from.day))
+        .inDays;
+  }
+
   @override
   bool operator ==(Object other) =>
       other is WeeklyWeightChange &&
       other.weekStart == weekStart &&
-      other.delta == delta &&
-      other.unit == unit &&
-      other.firstDate == firstDate &&
-      other.firstValue == firstValue &&
-      other.lastDate == lastDate &&
-      other.lastValue == lastValue;
+      _sameWeighIn(other.first, first) &&
+      _sameWeighIn(other.last, last);
 
   @override
   int get hashCode => Object.hash(
     weekStart,
-    delta,
-    unit,
-    firstDate,
-    firstValue,
-    lastDate,
-    lastValue,
+    first?.date,
+    first?.value,
+    last?.date,
+    last?.value,
   );
+
+  /// `Weight` compares by identity, so weigh-ins are matched on what they say.
+  static bool _sameWeighIn(Weight? a, Weight? b) =>
+      a?.date == b?.date && a?.value == b?.value && a?.unit == b?.unit;
 
   @override
   String toString() =>
       'WeeklyWeightChange(weekStart: $weekStart, delta: $delta, unit: $unit, '
-      'first: $firstValue on $firstDate, last: $lastValue on $lastDate)';
+      'first: ${first?.value} on ${first?.date}, '
+      'last: ${last?.value} on ${last?.date})';
 }
--- a/lib/features/weight/data/weight_analytics.dart
+++ b/lib/features/weight/data/weight_analytics.dart
@@ -47,19 +47,34 @@ class WeightAnalytics {
   }
 
   static WeeklyWeightChange _change(DateTime weekStart, List<Weight>? entries) {
-    if (entries == null || entries.length < WeeklyWeightChange.minEntries) {
+    if (entries == null || entries.isEmpty) {
       return WeeklyWeightChange(weekStart: weekStart);
     }
 
     entries.sort((a, b) => a.date.compareTo(b.date));
     final first = entries.first;
     final last = entries.last;
+    if (_daysBetween(first.date, last.date) < WeeklyWeightChange.minSpanDays) {
+      return WeeklyWeightChange(weekStart: weekStart);
+    }
+
     return WeeklyWeightChange(
       weekStart: weekStart,
       delta: last.value - first.value,
       unit: last.unit,
+      firstDate: first.date,
+      firstValue: first.value,
+      lastDate: last.date,
+      lastValue: last.value,
     );
   }
+
+  /// Calendar days from [from] to [to], counted on UTC midnights so a
+  /// daylight-saving shift inside the week cannot shorten the span.
+  static int _daysBetween(DateTime from, DateTime to) =>
+      DateTime.utc(to.year, to.month, to.day)
+          .difference(DateTime.utc(from.year, from.month, from.day))
+          .inDays;
 
   /// The Sunday opening [date]'s week, at local midnight. Dart counts Mon=1…
   /// Sun=7, so `% 7` maps Sunday to a zero offset.
--- a/lib/features/weight/data/weight_analytics.dart
+++ b/lib/features/weight/data/weight_analytics.dart
@@ -42,39 +42,13 @@ class WeightAnalytics {
     }
 
     return [
-      for (final start in weekStarts) _change(start, entriesByWeek[start]),
+      for (final start in weekStarts)
+        WeeklyWeightChange(
+          weekStart: start,
+          entries: entriesByWeek[start] ?? const [],
+        ),
     ];
   }
-
-  static WeeklyWeightChange _change(DateTime weekStart, List<Weight>? entries) {
-    if (entries == null || entries.isEmpty) {
-      return WeeklyWeightChange(weekStart: weekStart);
-    }
-
-    entries.sort((a, b) => a.date.compareTo(b.date));
-    final first = entries.first;
-    final last = entries.last;
-    if (_daysBetween(first.date, last.date) < WeeklyWeightChange.minSpanDays) {
-      return WeeklyWeightChange(weekStart: weekStart);
-    }
-
-    return WeeklyWeightChange(
-      weekStart: weekStart,
-      delta: last.value - first.value,
-      unit: last.unit,
-      firstDate: first.date,
-      firstValue: first.value,
-      lastDate: last.date,
-      lastValue: last.value,
-    );
-  }
-
-  /// Calendar days from [from] to [to], counted on UTC midnights so a
-  /// daylight-saving shift inside the week cannot shorten the span.
-  static int _daysBetween(DateTime from, DateTime to) =>
-      DateTime.utc(to.year, to.month, to.day)
-          .difference(DateTime.utc(from.year, from.month, from.day))
-          .inDays;
 
   /// The Sunday opening [date]'s week, at local midnight. Dart counts Mon=1…
   /// Sun=7, so `% 7` maps Sunday to a zero offset.
--- a/lib/ui/widgets/weekly_change_heatmap.dart
+++ b/lib/ui/widgets/weekly_change_heatmap.dart
@@ -1,11 +1,13 @@
 import 'package:flutter/material.dart';
 import 'package:food_locker/features/weight/data/weekly_weight_change.dart';
 import 'package:food_locker/ui/widgets/weekly_change_color.dart';
+import 'package:food_locker/ui/widgets/weekly_change_summary.dart';
 
 /// A year of weekly weight change as a grid of coloured cells.
 ///
 /// Fills row by row, oldest top-left to newest bottom-right, so the last cell
-/// is the week in progress. Unlabelled and non-interactive by design.
+/// is the week in progress. Holding a cell names the week it covers and the
+/// weigh-ins its colour came from.
 class WeeklyChangeHeatmap extends StatelessWidget {
   const WeeklyChangeHeatmap({super.key, required this.weeks});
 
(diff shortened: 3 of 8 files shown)


igorbasko01:
**Verdict: nothing blocks merge.** The gate swap is sound and strictly widening, the model addition is contained (no Hive `typeId`, no Drift schema, no backup codec touched), every `WeeklyWeightChange` construction site was updated to satisfy the new all-or-nothing assert, and every acceptance criterion in #115 has a test behind it. CI is green. What follows is one interaction consequence worth a look and three nits — none of them merge blockers.

### 1. A scroll that starts on the heatmap flashes a tooltip — `lib/ui/widgets/weekly_change_heatmap.dart:75-77`

`onPointerDown` calls `ensureTooltipVisible()` immediately, with no delay and no arena participation. `HomePage` is a `CustomScrollView` (`lib/ui/pages/home_page.dart:25`) and the heatmap is a full-width sliver near the top, so it is a natural place to start a scroll drag.

Concrete case: the user puts a finger on a cell and drags up to reach the history list. The tooltip pops up on touch-down, stays up for the whole drag (`ensureTooltipVisible` passes no `showDuration`, so nothing auto-hides it), and — because a `Tooltip` overlay is positioned once when shown and does not follow the scroll — hangs over the scrolling content at the cell's original screen position until the finger lifts. It does dismiss correctly on release (`onPointerUp` always reaches a `Listener`, arena outcome notwithstanding), so this is a flicker rather than a stuck overlay, but it fires on every scroll begun over the grid.

The `Tooltip` in `weekly_change_heatmap_test.dart` is only ever pressed in place, never dragged, so nothing catches this. Worth considering a short delay before showing, or dismissing once the pointer moves beyond the touch slop.

### 2. A week with no weigh-ins at all claims the weigh-ins were too close together — `lib/ui/widgets/weekly_change_summary.dart:19-23`

`weeklyChangeSummary` has one branch for `!week.hasData`, so a week that failed the span gate and a week the user never stepped on the scale in produce the same second line: `No weigh-ins at least 3 days apart`.

Concrete case: a user who logs sporadically holds a cell for a week they skipped entirely and reads a sentence implying they *did* weigh in, just not far enough apart. The model distinguishes the two cases only by absence, so telling them apart would need the entry count (or a plain `No weigh-ins this week`) — but as written the line is more specific than the data supports. #115 asked for a clear no-data state, and this is the one place it is slightly misleading.

### 3. Sub-0.05 deltas render as `+0.0` / `-0.0` — `lib/ui/widgets/weekly_change_summary.dart:41-48`

`_measurement` uses `toStringAsFixed(1)`, and `_signedChange` prefixes `+` on any `delta > 0`.

Concrete case: Sunday 82.44 kg, Saturday 82.48 kg. `level` is 1, so the cell draws faint red, and the tooltip's last line reads `+0.0 kg`. A 0.04 kg loss reads `-0.0 kg`, which looks like a formatting bug to a reader. This matches `WeightChangeIndicator` exactly, so it is a pre-existing convention rather than something this PR introduced — flagging it only because the tooltip now puts the rounded delta directly under the two endpoints it was derived from, where the mismatch is visible in one glance.

### 4. 52 summaries, rebuilt on every Home frame — `lib/ui/widgets/weekly_change_heatmap.dart:56`

`weeklyChangeSummary(week)` runs in `_CellState.build`, and each data week makes three `shortDateWithWeekday` calls, each constructing a fresh `DateFormat`. That is roughly 150 `DateFormat` constructions per `HomePage` rebuild, and `HomePage` rebuilds on every `WeightManager` notify — every weight added or deleted, every history-range change — even though the string is needed only while a cell is held (and, for the `Semantics` label, only when a screen reader is attached). `week` is immutable, so caching the summary in `_CellState` and recomputing it in `didUpdateWidget` when `widget.week` changes would take the whole cost out of the steady state.

### What checks out

- **Gate swap is not a regression.** With one entry per day, four weigh-ins in a Sun–Sat week always span at least three days, so every week that passed `minEntries = 4` still passes `minSpanDays = 3`. `home_page_test`'s 28-day and single-entry cases both keep their outcomes, and the `any(hasData)` visibility gate on the heatmap is unaffected.
- **Span arithmetic.** `_daysBetween` normalizes to UTC midnights of the local `(y, m, d)`, and `_weekEnd` steps by calendar components rather than `Duration(days: 7)` — both DST-safe, consistent with the existing `_weekStart` / `_weekStartsEndingAt`.
- **Model invariant.** The constructor assert makes the endpoints all-or-nothing alongside `delta`, `==`/`hashCode` were extended in step, and all seven other construction sites (three in `WeightAnalytics`, four in tests) were updated — nothing trips the assert.
- **Repo invariants.** No `@HiveType`/`@HiveField` change, no `typeId`, no `*.g.dart` touched, no Drift table or `schemaVersion`, no backup codec or `SerializationService` surface, no manual `pubspec.yaml` / `CHANGELOG.md` bump. Day granularity is untouched — the endpoints are read, never written.
- **Title.** `feat:` is right for a user-visible behaviour change; the fixed faint-green single-entry cell rides along as part of the gate rather than needing its own prefix.
- **Tests.** Both new files mirror `lib/`, analytics tests go through `InMemoryWeightRepository`, and the summary is asserted at both `en_US` and `en_GB` so the locale field order is actually exercised.

---

igorbasko01:
Thanks — acted on the two findings that are defects in this PR's own feature, and left the other two for you to weigh.

**Fixed in `013d6c0`:**

1. **Scroll drag popping a tooltip.** Real, and exactly the case the tests missed. `_CellState` now records the pointer-down position and dismisses the tooltip once the finger travels past `kTouchSlop`, so a drag that starts on the grid scrolls Home without leaving an overlay pinned at the cell's original position. Press-in-place is unchanged. New test: *a press that turns into a scroll drops the tooltip*.

2. **A skipped week claiming its weigh-ins were too close.** Also right — the line was more specific than the data supports. Both cases now read `No weigh-ins far enough apart to compare`, which is true whether the week held nothing or held two weigh-ins a day apart. Distinguishing them properly would mean carrying an entry count on the model, which is more than #115 asks for.

**Left as-is:**

3. **`+0.0` / `-0.0` for sub-0.05 deltas.** Pre-existing convention from `weight_change_indicator.dart`, and the tooltip deliberately matches it so a cell and the history rows below read the same way. Changing the rounding rule belongs in a change that changes it everywhere, not here.

4. **52 summaries rebuilt per `HomePage` frame.** Fair, and the `didUpdateWidget` cache would work — but this is an inefficiency on a rebuild driven by discrete user actions, not a per-frame path, and caching it widens the PR past the issue. Worth its own change if the Home rebuild ever shows up in a trace.

---

igorbasko01:
Please build a test apk

igorbasko01:
Test APK is built and published — **`1.18.0-pr131.12`**, from this PR's head `f4d42ed`.

📱 **[food_locker-test.apk](https://github.com/igorbasko01/food-locker/releases/download/test-builds/food_locker-test.apk)** (58 MB) — [release page](https://github.com/igorbasko01/food-locker/releases/tag/test-builds) · [build run](https://github.com/igorbasko01/food-locker/actions/runs/33780756473)

Ran the existing `Build Test APK` workflow with `label: pr131` (I have no Flutter or Android SDK in this session, and `CLAUDE.md` says not to install them). It installs side by side with production as **FoodLocker Test** with its own empty data — import a backup from the production app's Settings tab to try it against real weeks. As documented, this replaced the previous `test-builds` prerelease; the download URL is unchanged.

Worth exercising on the phone:

- **Hold a coloured cell** — it should name the week, the two weigh-ins compared, and the signed change, and disappear the moment you lift.
- **Hold a grey cell** — `No weigh-ins far enough apart to compare` rather than silence.
- **Start a scroll from the grid** — a drag past the touch slop should dismiss the tooltip instead of leaving it floating over the page. This one is the newest fix and the least covered by the widget tests.
- **A week you weighed only on Sunday and Saturday** should now be coloured instead of grey — that's the span gate. A week with a single weigh-in should be grey rather than faint green.

---

igorbasko01:
The tooltip should appear above the tile, otherwise the tooltip hides behind the finger, blocked out of view.

igorbasko01:
Good catch — fixed in `80a88d5`, and a fresh test APK is up.

The `Tooltip` defaulted to `preferBelow: true`, so it opened straight into the hand holding the cell. It now prefers the space **above** the cell, with `verticalOffset: 36` — the offset is measured from the cell's *centre*, and a cell is only a few millimetres across, so the default 24 would have left the tooltip almost touching the fingertip. Flutter still drops it below on its own for a cell too near the top of the screen to fit above.

A new widget test presses a cell in the bottom row and asserts the tooltip's rect ends above the cell's top edge, so this can't silently regress.

📱 **[food_locker-test.apk](https://github.com/igorbasko01/food-locker/releases/download/test-builds/food_locker-test.apk)** — `1.18.0-pr131.13`, from `80a88d5` · [build run](https://github.com/igorbasko01/food-locker/actions/runs/33785651542)

`analyze_and_test` is green on the new head. Worth a second look on the phone: a cell in the **top row** of the grid is the one case where Flutter will still place the tooltip below — if that reads badly in practice, the fix is to give the grid more headroom on Home rather than to fight the placement.

---

lib/ui/widgets/weekly_change_heatmap.dart
```diff
@@ -24,33 +27,91 @@ class WeeklyChangeHeatmap extends StatelessWidget {
...
-  final Color color;
+  /// The week this cell draws, or null past the end of the grid.
+  final WeeklyWeightChange? week;
+
+  @override
+  State<_Cell> createState() => _CellState();
+}
+
+class _CellState extends State<_Cell> {
+  final GlobalKey<TooltipState> _tooltip = GlobalKey<TooltipState>();
```
igorbasko01: Whats global key?

lib/ui/widgets/weekly_change_heatmap.dart
```diff
@@ -24,33 +27,91 @@ class WeeklyChangeHeatmap extends StatelessWidget {
...
+  Offset? _pressOrigin;
+
+  void _showTooltip(PointerDownEvent event) {
+    _pressOrigin = event.position;
+    _tooltip.currentState?.ensureTooltipVisible();
+  }
+
+  /// A finger that travels is scrolling Home, not reading a cell, and the
+  /// overlay stays where it opened rather than following the page.
+  void _dismissIfDragged(PointerMoveEvent event) {
```
igorbasko01: Shouldn't it move to next cell when dragging?

lib/ui/widgets/weekly_change_summary.dart
```diff
@@ -0,0 +1,47 @@
...
+List<String> weeklyChangeSummary(WeeklyWeightChange? week, [String? locale]) {
+  if (week == null) return const ['No data'];
+
+  final period =
+      '${shortDateWithWeekday(week.weekStart, locale)} – '
+      '${shortDateWithWeekday(_weekEnd(week.weekStart), locale)}';
+  // True of a week with nothing logged as much as one whose weigh-ins sat
+  // too close together.
+  if (!week.hasData) {
+    return [period, 'No weigh-ins far enough apart to compare'];
```
igorbasko01: Maybe tell the rule of the span?

lib/features/weight/data/weekly_weight_change.dart
```diff
@@ -4,22 +4,59 @@ import 'package:food_locker/features/weight/data/weight.dart';
...
 /// Intra-week only: [delta] is the week's last weigh-in minus its first, so
 /// nothing carries across a week boundary and every week stands on its own. A
-/// week under [minEntries] weigh-ins reports no [delta] at all.
+/// week whose weigh-ins span fewer than [minSpanDays] days reports no [delta]
+/// at all.
 class WeeklyWeightChange {
-  const WeeklyWeightChange({required this.weekStart, this.delta, this.unit});
+  const WeeklyWeightChange({
+    required this.weekStart,
+    this.delta,
```
igorbasko01: Why delta is part of the constructor and not calculated from first and last day?

lib/features/weight/data/weight_analytics.dart
```diff
@@ -47,20 +47,35 @@ class WeightAnalytics {
...
   static WeeklyWeightChange _change(DateTime weekStart, List<Weight>? entries) {
-    if (entries == null || entries.length < WeeklyWeightChange.minEntries) {
+    if (entries == null || entries.isEmpty) {
       return WeeklyWeightChange(weekStart: weekStart);
     }
 
     entries.sort((a, b) => a.date.compareTo(b.date));
     final first = entries.first;
     final last = entries.last;
+    if (_daysBetween(first.date, last.date) < WeeklyWeightChange.minSpanDays) {
```
igorbasko01: Why not return in any case, and the weeklyweightchange to handle the delta and span and everything?

lib/features/weight/data/weight_analytics.dart
```diff
@@ -47,20 +47,35 @@ class WeightAnalytics {
...
+      return WeeklyWeightChange(weekStart: weekStart);
+    }
+
     return WeeklyWeightChange(
       weekStart: weekStart,
       delta: last.value - first.value,
       unit: last.unit,
+      firstDate: first.date,
```
igorbasko01: Doesn't each weight holds info of date and weight? Cant we just pass a list of weights?

lib/features/weight/data/weekly_weight_change.dart
igorbasko01: Fair — it shouldn't have been. Changed in `1a88bab`: `delta` is a getter now.

```dart
double? get delta => hasData ? last!.value - first!.value : null;
```

The reason it was passed in was that `WeightAnalytics` had already computed it while picking the endpoints, so handing it over saved recomputing. That's a bad trade: it made the same fact storable in two places, which is exactly what the all-or-nothing constructor assert existed to police. With `delta`, `unit` and `hasData` all derived from `first` and `last`, that assert is gone and the state can't disagree with itself.

---

lib/features/weight/data/weight_analytics.dart
igorbasko01: Agreed — done in `1a88bab`. `_change` is gone entirely, and `weeklyChanges` now just buckets weigh-ins into weeks and constructs one per week:

```dart
return [
  for (final start in weekStarts)
    WeeklyWeightChange(
      weekStart: start,
      entries: entriesByWeek[start] ?? const [],
    ),
];
```

The span gate, the delta and the unit all live in `WeeklyWeightChange` now, so there's one place that decides what a week's change is. `WeightAnalytics` is left with the job only it can do — reading the repository and working out which week each weigh-in belongs to. It also lost `_daysBetween`, which moved to the model as `_spanInDays`.

---

lib/features/weight/data/weight_analytics.dart
igorbasko01: It does, and yes — that's what `1a88bab` now does. The model takes the week's weigh-ins and keeps two `Weight`s rather than four unpacked fields:

```dart
factory WeeklyWeightChange({
  required DateTime weekStart,
  List<Weight> entries = const [],
})
```

My original reason for unpacking them was narrow: `Weight` has no `==`, so holding `Weight` objects would have made `WeeklyWeightChange`'s equality quietly identity-based, and two weeks built from equal weigh-ins would have compared unequal. That's solved locally instead of by flattening the model — `==` compares the weigh-ins on date, value and unit:

```dart
/// `Weight` compares by identity, so weigh-ins are matched on what they say.
static bool _sameWeighIn(Weight? a, Weight? b) =>
    a?.date == b?.date && a?.value == b?.value && a?.unit == b?.unit;
```

Giving `Weight` itself value equality would be tidier, but it's a Hive model used well outside this feature and changing how it compares could shift behaviour in code this PR has no business touching — worth doing separately if you want it.

One small bonus: the factory scans for the earliest and latest weigh-in instead of sorting, so it no longer sorts the caller's list in place.

---

lib/ui/widgets/weekly_change_summary.dart
igorbasko01: Done in `1a88bab` — the line now names the rule instead of gesturing at it:

> `No two weigh-ins 3 days apart`

with the `3` interpolated from `WeeklyWeightChange.minSpanDays`, so the text can't drift from the gate if the constant ever changes.

Worth recording why it isn't phrased as "your weigh-ins were less than 3 days apart": the same line has to serve a week you never stepped on the scale in. The earlier wording (`No weigh-ins at least 3 days apart`) implied you *had* weighed in, just too close together, which is wrong for an empty week. "No two weigh-ins 3 days apart" is true either way.

---

lib/ui/widgets/weekly_change_heatmap.dart
igorbasko01: A `GlobalKey` is a key that's unique across the whole widget tree, and its useful property here is that it gives you a handle on the *state object* of the widget it's attached to — `_tooltip.currentState` is the live `TooltipState`. Normal keys only help Flutter match widgets to elements when a list reorders; a `GlobalKey` additionally lets code outside the widget reach in.

I need that because the tooltip is driven by hand. `Tooltip` has no "shown" property to rebuild with — showing it means calling `ensureTooltipVisible()` on its state — so `triggerMode: TooltipTriggerMode.manual` plus the key is how the `Listener` opens it on pointer-down and closes it on pointer-up.

I've added a comment saying so in `fa84f6d`:

```dart
/// Reaches the [Tooltip]'s own state to show it on demand, which is the only
/// way in: the widget exposes no "shown" flag to rebuild with.
final GlobalKey<TooltipState> _tooltip = GlobalKey<TooltipState>();
```

Worth knowing they aren't free: a `GlobalKey` makes Flutter track that element globally, and moving one around the tree forces a rebuild of its subtree. One per cell (52 of them) is cheap because they never move — but they're not something to reach for by default.

---

lib/ui/widgets/weekly_change_heatmap.dart
igorbasko01: It'd be nicer, yes — dragging across the grid to read week after week is the natural gesture for a heatmap. I haven't built it here, because it isn't a tweak to this line: it's a different interaction model, and it collides with the one thing this drag handler exists to protect.

The problem is that each cell currently owns its own gesture. A `Listener` only receives moves for the pointer that went down on *it*, so a cell can't hand the press to its neighbour. Scrubbing needs the whole grid to own the gesture: one recognizer above the 52 cells, hit-testing the drag position to a cell index on every move and driving that cell's tooltip. That means the tooltips can no longer be per-cell `Tooltip` widgets driven by a `GlobalKey` — the grid would hold a single overlay it repositions.

The collision: the heatmap is a sliver in Home's `CustomScrollView`, and a drag that starts on the grid is currently how you scroll past it — which is exactly what I made the touch-slop dismissal do, after you found the tooltip hanging over the page. Claim vertical drags for scrubbing and scrolling from the grid stops working; claim only horizontal ones and scrubbing works along a row but dies the moment your finger drifts up or down, which on a 4-row grid of ~5 mm cells it will.

So there's a real decision to make (horizontal-only scrub? long-press to arm, then scrub freely, as iOS charts do?), and it wants trying on a phone rather than guessing. I'd rather not fold that into #115 — happy to open it as its own issue with the above written up, if you want it.

---

igorbasko01:
All six comments answered — four led to changes, two are answers in their threads.

**Changed (`1a88bab`, `fa84f6d`):**

| Your comment | What changed |
|---|---|
| Why is `delta` in the constructor? | It's a getter now, derived from the two weigh-ins |
| Why not always return, and let `WeeklyWeightChange` handle it? | `_change` is gone; `weeklyChanges` constructs one per week and the model owns the span gate |
| Can't we just pass a list of weights? | It takes `List<Weight>` and keeps `first`/`last` as `Weight`s |
| Tell the rule of the span | `No two weigh-ins 3 days apart`, with the `3` from `minSpanDays` |

Those three model comments were really one point, and following it paid off: `WeightAnalytics` lost ~30 lines and is back to only bucketing weigh-ins into weeks, and the all-or-nothing constructor assert is gone because there's no longer duplicated state to police. Net −8 lines across the PR with the tests updated.

**Answered, not changed:**

- **What's a `GlobalKey`** — explained in the thread, and `fa84f6d` adds a comment at the declaration so the next reader doesn't have to ask.
- **Should dragging move to the next cell** — it'd be nicer, but it's a different interaction model (one recognizer above the grid instead of per-cell gestures) and it fights the scroll-drag dismissal you asked for earlier. Reasoning is in the thread; happy to open it as its own issue.

`analyze_and_test` is green on `fa84f6d`, and a fresh APK is up with the new tooltip wording:

📱 **[food_locker-test.apk](https://github.com/igorbasko01/food-locker/releases/download/test-builds/food_locker-test.apk)** — `1.18.0-pr131.14` · [build run](https://github.com/igorbasko01/food-locker/actions/runs/33964343048)

---

lib/ui/widgets/weekly_change_heatmap.dart
igorbasko01: That is why it sounds strange to me to use it. I assume that in the weights and bites chart, there is nothing similar, but still shows a tooltip. And also it is draggable on those charts.

lib/features/weight/data/weekly_weight_change.dart
```diff
@@ -2,25 +2,58 @@ import 'package:food_locker/features/weight/data/weight.dart';
...
+/// Built from the week's own weigh-ins and derives everything from them:
+/// intra-week only, so nothing carries across a week boundary and every week
+/// stands on its own. A week whose weigh-ins span fewer than [minSpanDays]
+/// days reports no [delta] at all.
 class WeeklyWeightChange {
-  const WeeklyWeightChange({required this.weekStart, this.delta, this.unit});
+  /// The week [weekStart] opens, measured across [entries] — the week's
+  /// weigh-ins, in any order. An empty list is a week nothing was logged in.
+  factory WeeklyWeightChange({
+    required DateTime weekStart,
```
igorbasko01, line 13: Do we need the weekstart if we pass all the weights?

lib/ui/widgets/weekly_change_heatmap.dart
igorbasko01: shouldn't the heatmap be a self contained widget? so it could handle gestures as it sees fit?

igorbasko01: Yes it should, and you were right to keep pushing — that's `fa13ba8`. The heatmap now owns its gesture and draws its own readout, so dragging carries it from cell to cell.

What changed:

- One `Listener` over the whole grid instead of 52 per-cell ones. `LayoutBuilder` gives the cell size, so the cell under the finger is arithmetic — `local.dx ~/ cellSize`, `local.dy ~/ cellSize` — not widget hit-testing.
- No `Tooltip` widgets and no `GlobalKey`s at all. The grid keeps one `OverlayEntry` and rebuilds it as the finger moves.
- Down shows the cell's readout, move switches it to whichever cell you're now over, up or leaving the grid ends it.

The scroll collision I described turned out not to bite. Because the cell is recomputed from the finger's position on every move, a drag that scrolls the page keeps the readout glued to the cell under the finger instead of leaving it stranded — so the touch-slop dismissal I added earlier is gone, and scrolling from the grid still works. That's strictly better than either option I laid out before; I was reasoning about a fixed overlay, and once it follows the finger the problem dissolves.

The charts get this from `fl_chart` owning its whole canvas and painting tooltips into it. This is the same shape, hand-rolled, because our cells are 52 real widgets rather than one painted surface.

Tests cover the drag now: *dragging carries the readout to the next cell*, *dragging off the grid ends the readout*, plus the existing press/release and above-the-finger placement.

---

lib/ui/widgets/weekly_change_heatmap.dart
igorbasko01: You were right to find it strange. The charts don't need a `GlobalKey` because `fl_chart` owns its entire canvas: one widget handles the touch and paints the tooltip itself, so there's no separate widget whose state anything needs to reach into. My heatmap had it backwards — 52 cells each owning a `Tooltip`, which then needed a key each to be driven from outside.

`fa13ba8` follows the chart's shape instead: the grid owns the gesture and draws its own readout, and the keys are gone entirely. Details in the [other thread](https://github.com/igorbasko01/food-locker/pull/131#discussion_r3944634777) — it's the same change that gives you drag-between-cells.

The explanation of what a `GlobalKey` is still stands as background, but it no longer describes this file.

---

lib/features/weight/data/weekly_weight_change.dart
igorbasko01: Yes — a week with **no** weigh-ins still has to know which week it is. The heatmap draws all 52 cells whether or not the store has anything for them, and holding an empty one names its period (`Sun, 3/8 – Sat, 3/14`). Derive `weekStart` from the entries and an empty week has nothing to derive it from, so the grid couldn't say what the grey cell covers.

Two smaller reasons it stays: deriving it would duplicate the Sunday-boundary rule that `WeightAnalytics._weekStart` already owns, and it would quietly accept entries from two different weeks and produce a nonsense object.

That last one was worth closing properly, so `e15bc42` makes the relationship explicit instead of implied:

```dart
assert(
  entries.every((entry) => _fallsInWeek(entry, weekStart)),
  'every weigh-in must fall in the week weekStart opens',
);
```

with a test that a weigh-in from the following week is rejected. So `weekStart` is the week's identity and `entries` are what happened in it — the assert is what keeps those two from disagreeing.

---

igorbasko01:
Test APK for the reworked interaction is up — **`1.18.0-pr131.15`**, built from `e15bc42`.

📱 **[food_locker-test.apk](https://github.com/igorbasko01/food-locker/releases/download/test-builds/food_locker-test.apk)** · [build run](https://github.com/igorbasko01/food-locker/actions/runs/34047158533)

I see `main` has since been merged in (`0355961`), which brought #132's release workflow and `RELEASING.md` — no `lib/` or `test/` change, so the APK above is still a faithful build of the app code on the current head. `analyze_and_test` is green on `0355961` and the PR is mergeable.

Worth trying on the phone, since this round is all interaction:

- **Drag across the grid** — the readout should hand off cell to cell as your finger crosses them, and keep up if the page scrolls under you.
- **Lift, or drag off the bottom of the grid** — it should end either way.
- **Press a top-row cell** — the one case where the readout has no room above and drops below instead.

Everything from the review round is answered: the `GlobalKey`s are gone, the grid owns its gesture, `weekStart` earns its place with an assert holding entries to their week, and the PR description now describes the design as it actually stands rather than the two revisions it went through.

---

PR #113, opened 2026-08-29
chore(main): release 1.19.0

igorbasko01:
:robot: I have created a release *beep* *boop*
---

### Features

* add a 52-week weight-change heatmap to Home ([#114](https://github.com/igorbasko01/food-locker/issues/114)) ([55ba43f](https://github.com/igorbasko01/food-locker/commit/55ba43f35d68a49e738ae972457cf2a639d0dc0d))
* capture the user's height in Settings, in metric or imperial ([#133](https://github.com/igorbasko01/food-locker/issues/133)) ([6b9fefc](https://github.com/igorbasko01/food-locker/commit/6b9fefc7acb47aab223356ea7da2bd4d18ff4b6d))
* explain the Home heatmap cells and gate weeks on span ([#131](https://github.com/igorbasko01/food-locker/issues/131)) ([5fcd67b](https://github.com/igorbasko01/food-locker/commit/5fcd67b467c37b2bddfe6ab95250de9f101774e2))
* remove the Home streak banners and overeating analytics ([#112](https://github.com/igorbasko01/food-locker/issues/112)) ([5549abd](https://github.com/igorbasko01/food-locker/commit/5549abd3c05d830fb2de0d774ce2a594338c4e70))
* show the weekday alongside every displayed date ([#116](https://github.com/igorbasko01/food-locker/issues/116)) ([#128](https://github.com/igorbasko01/food-locker/issues/128)) ([4a745f1](https://github.com/igorbasko01/food-locker/commit/4a745f1184804163484bda13f1154c2d42148234))
* weight tab stat tiles show change, not three flavours of lowest weight ([#134](https://github.com/igorbasko01/food-locker/issues/134)) ([31150af](https://github.com/igorbasko01/food-locker/commit/31150af09ed6319a32e2f7fe0734f3b77921311b))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #132, opened 2026-09-04
chore: merge the green release-please PR every Sunday

### LINKED ISSUE #118: chore: auto-merge the open release-please PR every Sunday when checks are green
## Goal

Every Sunday, if there is an open release-please PR and all of its checks have passed, merge it automatically so a release and a production APK are produced without manual intervention.

## Background

The post-merge half of this is already automatic:

1. Merging the release PR pushes to `main`.
2. `.github/workflows/release-please.yml` runs and cuts the GitHub Release + tag.
3. `.github/workflows/build_release.yml` fires on `release: published`, builds the signed APK, and uploads it to the release.

So the only missing link is the merge itself.

Two facts about the current setup that the implementation depends on:

- `release-please.yml` authenticates with `RELEASE_PLEASE_TOKEN` (a PAT), **not** `GITHUB_TOKEN`. That is why the release PR actually gets CI — e.g. #113 has `analyze_and_test` green.
- **The merge must therefore also use `RELEASE_PLEASE_TOKEN`.** A merge performed with `GITHUB_TOKEN` pushes to `main` without triggering `release-please.yml`, so no release is cut and no APK is built. This is the easiest way to get a silently broken automation.

The open release PR is identifiable by any of: the label `autorelease: pending`, the head branch prefix `release-please--branches--main--components--`, or the title `chore(main): release X.Y.Z`.

## Proposal

Add `.github/workflows/sunday-release-merge.yml`:

- Triggers: `schedule` (weekly, Sunday) and `workflow_dispatch` (a manual "ship it now" button, usable from the GitHub mobile app).
- Steps:
  1. Find the open PR labelled `autorelease: pending` targeting `main`. Exit cleanly if there is none.
  2. Read its check runs and mergeable state.
  3. Squash-merge with `RELEASE_PLEASE_TOKEN`, keeping the `chore(main): release X.Y.Z` title so release-please's `autorelease:` bookkeeping stays intact.
  4. Write a job summary in every path (merged / skipped and why), so a skipped week is visible on Monday.

### Schedule

`cron: '0 1 * * 0'` — 01:00 UTC Sunday, which is **04:00 Sunday in Israel during IDT** (UTC+3) and 03:00 during IST (UTC+2).

GitHub Actions evaluates cron in UTC with no DST handling, so a single expression cannot hold 04:00 local year-round. The one-hour winter drift is accepted: this is an unattended early-morning job, and GitHub's scheduled-run queue can delay a trigger by tens of minutes anyway, so the exact minute is not meaningful. (If exact 04:00 local is ever wanted, the alternative is two cron entries — `0 1 * * 0` and `0 2 * * 0` — with a first step that exits unless `TZ=Asia/Jerusalem date +%H` is `04`.)

Neither UTC time crosses midnight, so the cron day-of-week stays `0` (Sunday) in both seasons.

### Fail-closed rules

- Skip if any check run is `pending`, `failure`, `cancelled`, or `timed_out`.
- Skip if the PR reports **zero** check runs — this guards against a future config change that stops CI from running on the release PR, which would otherwise make the "all checks passed" gate vacuously true.
- Skip if the PR is not mergeable (conflict, draft).
- **Skip if more than one open PR carries `autorelease: pending`.** This should never happen; if it does, merge nothing and say so in the job summary rather than guessing which one is the real release.

## Alternatives considered

- **GitHub native auto-merge** (`gh pr merge --auto` armed weekly, plus branch protection requiring `analyze_and_test`). Lets GitHub enforce the green-checks gate with no polling code, and correctly handles checks that are still running at cron time. Rejected because auto-merge stays armed until it fires: if a Sunday's checks never go green, the next release-please force-push reruns checks and the PR merges mid-week, defeating the "Sunday only" intent — unless a separate disarm job is added.
- **Scheduled Claude session.** Adds judgement, but it is a purely mechanical gate; non-deterministic, invisible in the repo, and costs a run every week.
- **Manual `workflow_dispatch` only.** No unattended release, but not the requested behaviour. Included inside this proposal as a secondary trigger anyway.

## Acceptance criteria

- [ ] Workflow exists with both `schedule` (`0 1 * * 0`) and `workflow_dispatch` triggers.
- [ ] Merge is performed with `RELEASE_PLEASE_TOKEN`, so `release-please.yml` runs afterwards and `build_release.yml` produces the APK.
- [ ] A red, pending, unmergeable, or check-less release PR is left untouched, with the reason in the job summary.
- [ ] More than one open `autorelease: pending` PR merges nothing, with the reason in the job summary.
- [ ] No open release PR is a clean no-op.
- [ ] Verified end to end at least once via `workflow_dispatch`: release PR merged → release tagged → APK attached to the release.

igorbasko01:
## What changed

A new `.github/workflows/sunday-release-merge.yml` closes the last manual link in the release chain. It runs on `schedule` (`0 1 * * 0`) and on `workflow_dispatch`, finds the open PR labelled `autorelease: pending` targeting `main`, and squash-merges it once it is mergeable and every check run on its head commit is green. Everything after that is already automatic: the push to `main` triggers `release-please.yml`, which cuts the release and tag, which triggers `build_release.yml`, which attaches the APK.

The merge authenticates with `RELEASE_PLEASE_TOKEN`, not `GITHUB_TOKEN` — a push made with `GITHUB_TOKEN` does not trigger other workflows, so a merge performed with it would silently produce no release and no APK. The workflow also fails with a loud job summary if that secret is missing, rather than falling back to a token that would break the chain.

`RELEASING.md` §3 said "merge the bot's PR" as a manual step, so it now carries a short subsection describing the automatic merge, its manual **Run workflow** button, and the token requirement.

### Every path writes a job summary

| Situation | Outcome |
|---|---|
| No `autorelease: pending` PR | Clean no-op, "Nothing to release" |
| More than one such PR | Merges nothing, lists both, asks for a hand merge |
| PR is a draft | Skipped, reason recorded |
| PR is `CONFLICTING` (or still `UNKNOWN` after 5 retries) | Skipped, the reported state named |
| Head commit has **zero** check runs | Skipped — a vacuous "all green" is not a pass |
| Any check pending / failed / cancelled / timed out | Skipped, with each blocking check and its state listed |
| All green | Squash-merged, keeping the `chore(main): release X.Y.Z` subject |

## How each acceptance criterion is met

| From #118 | Where |
|---|---|
| Both `schedule` (`0 1 * * 0`) and `workflow_dispatch` triggers | `on:` block |
| Merge uses `RELEASE_PLEASE_TOKEN` | `GH_TOKEN: ${{ secrets.RELEASE_PLEASE_TOKEN }}`, the only credential in the job; `permissions: {}` leaves `GITHUB_TOKEN` with no scopes so it cannot be used by accident |
| Red / pending / unmergeable / check-less PR left untouched, reason in the summary | the guards above, each `exit 0` after a `note` |
| More than one `autorelease: pending` PR merges nothing, with the reason | `count -gt 1` branch, which lists the candidates |
| No open release PR is a clean no-op | `count -eq 0` branch, exit 0 |
| Verified end to end via `workflow_dispatch` | **Not done here** — see below |

- **`gh` CLI over the raw API**, matching `build_release.yml` and `build_test.yml`, which already shell out to `gh` on the runner.
- **Passing conclusions are an allowlist, not a denylist.** The issue names `pending`, `failure`, `cancelled` and `timed_out` as blocking; the script instead treats only `success`, `neutral` and `skipped` as passing, so a conclusion nobody anticipated (`action_required`, `stale`) blocks rather than slips through. Same intent, one fewer way to be wrong later.
- **Mergeability is retried before it is believed.** GitHub reports `UNKNOWN` while it recomputes the merge commit after a push, so the script polls five times at 10s before treating an unresolved state as a skip — fail-closed, without skipping a perfectly mergeable PR just because the API had not caught up.
- **The squash body is set to empty** (`--body ""`). `gh`'s default body is the PR's commit messages, and the release PR's changelog can legitimately contain a `BREAKING CHANGE:` line; leaving it in the squash commit on `main` risks release-please reading it as a footer and mis-bumping the *next* version.
- **The branch is not deleted** after merging. release-please force-pushes the same branch name for the next release, and deleting it here would add a race for no gain.
- **The winter DST hour of drift is accepted**, as the issue proposes — no second cron entry and no `TZ` guard step.

## Verification

This change is a workflow file, so there is nothing for `flutter analyze` or `flutter test` to cover, and no Dart test can exercise it. In this session neither local shell execution nor Python was available to lint the YAML or dry-run the script against a stubbed `gh`, so **the script was verified by review only** — read line by line against the `gh`/`jq` behaviours it depends on, and against the two existing workflows that already use `gh` on the runner.

Which leaves the issue's last acceptance criterion — *verified end to end at least once via `workflow_dispatch`: release PR merged → release tagged → APK attached* — deliberately **not** met by this PR. Doing it means merging the real open release PR (#113) and cutting a real release, which is not something to do unattended. Once this is on `main`, running **Sunday Release Merge → Run workflow** performs exactly that end-to-end check; #113 being green makes it a good first subject.

Worth knowing before that first run: `schedule` triggers only fire from the default branch, so nothing happens on any Sunday until this is merged.

Closes #118

https://claude.ai/code/session_01C7zZsPKbq5RebVocjJX18s

---
_Generated by [Claude Code](https://claude.ai/code/session_01C7zZsPKbq5RebVocjJX18s)_

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -37,6 +37,14 @@ When the team decides it is time to cut a release:
 3. The creation of that GitHub Release then triggers our `build_release.yml` workflow.
 4. `build_release.yml` builds a production-signed Android APK securely on CI, names it with the new version and build number, and uploads it to the action artifacts.
 
+### Merging the Release PR automatically
+
+Step 1 also happens on its own. `sunday-release-merge.yml` runs at 01:00 UTC every Sunday (04:00 in Israel during IDT) and squash-merges the open release PR, so a week's worth of merged features ships without anyone doing anything. The same workflow has a **Run workflow** button for shipping immediately, which works from the GitHub mobile app.
+
+It fails closed and merges nothing if the release PR is red, still running, has no checks at all, or cannot be merged — and if more than one PR carries the `autorelease: pending` label, since there is then no way to tell which one is the release. Every run writes what it did, or why it did nothing, to its job summary.
+
+The merge is performed with `RELEASE_PLEASE_TOKEN` rather than the default `GITHUB_TOKEN`: a push made with `GITHUB_TOKEN` does not trigger other workflows, so the release-please run that cuts the release and tag would never fire.
+
 ## 4. Setting up Release Signing
 
 To build a production release, you need a Keystore to cryptographically sign the Android APK.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -39,7 +39,7 @@ When the team decides it is time to cut a release:
 
 ### Merging the Release PR automatically
 
-Step 1 also happens on its own. `sunday-release-merge.yml` runs at 01:00 UTC every Sunday (04:00 in Israel during IDT) and squash-merges the open release PR, so a week's worth of merged features ships without anyone doing anything. The same workflow has a **Run workflow** button for shipping immediately, which works from the GitHub mobile app.
+Step 1 also happens on its own. `sunday-release-merge.yml` runs at 01:00 UTC every Sunday and squash-merges the open release PR, so a week's worth of merged features ships without anyone doing anything. The same workflow has a **Run workflow** button for shipping immediately, which works from the GitHub mobile app.
 
 It fails closed and merges nothing if the release PR is red, still running, has no checks at all, or cannot be merged — and if more than one PR carries the `autorelease: pending` label, since there is then no way to tell which one is the release. Every run writes what it did, or why it did nothing, to its job summary.
 
--- a/.github/workflows/sunday-release-merge.yml
+++ b/.github/workflows/sunday-release-merge.yml
@@ -80,7 +80,9 @@ jobs:
 
         # GitHub reports UNKNOWN while it recomputes mergeability after a push.
         for _ in 1 2 3 4 5; do
-          [ "$mergeable" != "UNKNOWN" ] && break
+          if [ "$mergeable" != "UNKNOWN" ]; then
+            break
+          fi
           sleep 10
           mergeable=$(gh pr view "$number" --repo "$REPO" --json mergeable --jq '.mergeable')
         done
@@ -101,7 +103,7 @@ jobs:
         fi
 
         blocking=""
-        while IFS="$(printf '\t')" read -r name status conclusion; do
+        while IFS=$'\t' read -r name status conclusion; do
           if [ "$status" != "completed" ]; then
             blocking="$blocking- \`$name\`: $status"$'\n'
           else
--- a/.github/workflows/sunday-release-merge.yml
+++ b/.github/workflows/sunday-release-merge.yml
@@ -7,9 +7,9 @@ name: Sunday Release Merge
 
 on:
   schedule:
-    # 01:00 UTC Sunday - 04:00 in Israel during IDT, 03:00 during IST. Actions
-    # evaluates cron in UTC with no DST handling, and delays scheduled runs
-    # anyway, so the winter hour of drift is accepted.
+    # Early Sunday. Actions evaluates cron in UTC with no DST handling, so the
+    # local hour this lands on drifts by one across the year; scheduled runs are
+    # queued and delayed anyway, so the drift is accepted.
     - cron: '0 1 * * 0'
   workflow_dispatch:
 
--- a/.github/workflows/sunday-release-merge.yml
+++ b/.github/workflows/sunday-release-merge.yml
@@ -122,7 +122,15 @@ jobs:
         fi
 
         # The squash subject has to stay `chore(main): release X.Y.Z` so
-        # release-please recognises its own release commit on `main`.
-        gh pr merge "$number" --repo "$REPO" --squash --subject "$title" --body ""
+        # release-please recognises its own release commit on `main`, and
+        # --match-head-commit refuses the merge if release-please force-pushed
+        # after the checks above were read - otherwise the green checks that
+        # let the merge through could belong to a commit that is no longer the
+        # one being merged.
+        if ! gh pr merge "$number" --repo "$REPO" --squash \
+          --match-head-commit "$sha" --subject "$title" --body ""; then
+          note "Not merged: the merge call failed - see the step log. The head commit may have moved since its checks were read."
+          exit 1
+        fi
 
-        note "Merged. release-please will tag the release, and build_release.yml will attach the APK to it."
+        note "Merged \`${sha:0:7}\`. release-please will tag the release, and build_release.yml will attach the APK to it."
--- a/.github/workflows/sunday-release-merge.yml
+++ b/.github/workflows/sunday-release-merge.yml
@@ -1,0 +1,126 @@
+name: Sunday Release Merge
+
+# Merges the open release-please PR on Sunday mornings once its checks are
+# green, so the release and its APK are cut without anyone at a keyboard.
+# Everything downstream is already automatic: the merge pushes to `main`,
+# release-please.yml tags the release, build_release.yml attaches the APK.
+
+on:
+  schedule:
+    # 01:00 UTC Sunday - 04:00 in Israel during IDT, 03:00 during IST. Actions
+    # evaluates cron in UTC with no DST handling, and delays scheduled runs
+    # anyway, so the winter hour of drift is accepted.
+    - cron: '0 1 * * 0'
+  workflow_dispatch:
+
+jobs:
+  merge:
+    runs-on: ubuntu-latest
+    # Every call below authenticates with RELEASE_PLEASE_TOKEN, so GITHUB_TOKEN
+    # needs no scopes of its own.
+    permissions: {}
+
+    steps:
+    - name: Merge the release PR when it is green
+      env:
+        # A merge performed with GITHUB_TOKEN pushes to `main` without
+        # triggering release-please.yml, which would leave no release and no
+        # APK - the same PAT that opens the release PR has to close it.
+        GH_TOKEN: ${{ secrets.RELEASE_PLEASE_TOKEN }}
+        REPO: ${{ github.repository }}
+      run: |
+        set -euo pipefail
+
+        note() { printf '%s\n' "$1" >> "$GITHUB_STEP_SUMMARY"; }
+
+        if [ -z "${GH_TOKEN:-}" ]; then
+          note "### Release merge failed"
+          note ""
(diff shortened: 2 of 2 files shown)


igorbasko01:
**Nothing blocks merge.** The workflow is correct as written, the token reasoning is right, and every branch does what the PR body says it does. Two minor robustness points and one process note below.

Since the authoring session had no shell, I extracted the `run:` script and executed it against a stubbed `gh`/`jq` locally. All eight documented paths reproduce exactly as described: green → one `gh pr merge 113 --repo … --squash --subject 'chore(main): release 1.19.0' --body ''`; red/pending → both blocking checks listed by name and state, no merge; zero check runs → skip; no PR → "Nothing to release"; two PRs → both listed, no merge; draft → skip; `CONFLICTING` → skip; `UNKNOWN` that stays unknown → skip after the retries, `UNKNOWN` that resolves → merge; empty `GH_TOKEN` → `exit 1` with the loud summary. The YAML parses. `bash -n` is clean, and the `while … done <<<"$checks"` loop is a herestring rather than a pipe, so `blocking` survives into the `if` — the usual subshell trap is avoided.

None of the repo's Dart invariants are in play: no `lib/`, no `*.g.dart`, no Hive `typeId` or `@HiveField` index, no Drift table or `schemaVersion`, no backup codec, and `pubspec.yaml`/`CHANGELOG.md` are untouched. The `chore:` prefix is right — CI plumbing plus docs, no user-visible app behaviour, so no version bump is wanted.

### 1. The merge is not pinned to the SHA whose checks were read (minor)

`sunday-release-merge.yml:68` captures `headRefOid` and `:95` validates check runs against exactly that commit, but `:126` merges whatever the PR's head happens to be at merge time.

The failure case is narrow but real: something lands on `main` shortly before 01:00 UTC Sunday, and release-please force-pushes the release branch in the window between the check read and `gh pr merge`. The green checks belonged to the old head; the new head's CI has not started, and the workflow merges it anyway. The gate is then vacuous in precisely the situation the zero-checks guard was written to prevent.

`gh pr merge` has `--match-head-commit` for this, and the SHA is already in hand:

```
gh pr merge "$number" --repo "$REPO" --squash --subject "$title" --body "" \
  --match-head-commit "$sha"
```

The call then fails loudly instead of merging an unverified commit, and the next Sunday (or a dispatch) picks it up.

### 2. A failed `gh pr merge` leaves the summary without an outcome line

`:126` runs under `set -e` with nothing after it in that path. If the merge call fails — a transient 5xx, a hand merge that raced it, a future branch-protection rule — the step aborts and the summary ends at "### Release PR #N" and the title, with no line saying what happened. The job is red so it is not silent, but it is the one path that does not honour the "every path writes a job summary" property the rest of the script maintains carefully. A trailing `|| { note "Not merged: the merge call failed, see the step log."; exit 1; }` would close it.

### 3. `Closes #118` ticks a box that is not ticked

The last acceptance criterion on #118 — verified end to end via `workflow_dispatch`, release PR merged → tag → APK attached — is explicitly not met here, and the PR is upfront about why. But `Closes #118` will auto-close the issue on merge with that criterion outstanding. Worth either dropping to a plain reference and closing #118 after the first dispatch run, or merging as-is and remembering that the verification is still owed. #113 (`chore(main): release 1.19.0`, `autorelease: pending`, `mergeable_state: clean`, `analyze_and_test` green) is a good first subject, and I confirmed it is currently the only PR carrying that label — so the single-candidate path is the one the first run will take.

### Checked and fine

- `permissions: {}` does not interfere with `$GITHUB_STEP_SUMMARY` (a file path, not a permission-gated API), and `GH_TOKEN` is the only credential the step uses.
- `${GH_TOKEN:-}` is the right guard under `set -u`: a missing secret arrives as an empty string, not as an unset variable, and the guard catches it.
- The `filter` parameter of `commits/{sha}/check-runs` defaults to `latest`, so a re-run does not surface a stale failing attempt alongside the passing one.
- Treating `success|neutral|skipped` as the passing allowlist rather than denylisting known-bad conclusions is the right call, as the PR body argues.
- `0 1 * * 0` is 04:00 IDT / 03:00 IST, and neither crosses midnight, so the day-of-week stays Sunday in both seasons — the comment at `:9` is accurate.
- Not deleting the branch is correct; release-please force-pushes that same name for the next release.

---

igorbasko01:
Thanks — the stubbed-`gh` dry run is exactly the verification this session couldn't run, and both mechanical findings were real. Fixed in `cee9413`:

**1. Merge pinned to the verified SHA.** The merge now passes `--match-head-commit "$sha"`, the same commit whose check runs were read, so a release-please force-push landing between the check read and the merge makes GitHub refuse the merge rather than land a commit whose CI never ran.

**2. A failed merge now summarizes.** `gh pr merge` is wrapped in an `if !`, writing a "Not merged: the merge call failed" line (naming the moved-head case, now the likeliest cause) before `exit 1`. Every path writes an outcome again. The success line also names the merged short SHA.

**3. `Closes #118` with the last acceptance criterion unmet** — left as is, deliberately. The end-to-end check means merging the real release PR (#113) and cutting a release, which isn't something to do unattended, and the PR body says so. Merging this will close the issue with that one box unticked; if you'd rather keep it open until the first `workflow_dispatch` run proves the chain, drop the `Closes` line when you squash and close it by hand afterwards.

---

RELEASING.md
```diff
@@ -37,6 +37,14 @@ When the team decides it is time to cut a release:
 3. The creation of that GitHub Release then triggers our `build_release.yml` workflow.
 4. `build_release.yml` builds a production-signed Android APK securely on CI, names it with the new version and build number, and uploads it to the action artifacts.
 
+### Merging the Release PR automatically
+
+Step 1 also happens on its own. `sunday-release-merge.yml` runs at 01:00 UTC every Sunday (04:00 in Israel during IDT) and squash-merges the open release PR, so a week's worth of merged features ships without anyone doing anything. The same workflow has a **Run workflow** button for shipping immediately, which works from the GitHub mobile app.
```
igorbasko01: No need to mention israel in any of the files.

igorbasko01: Done in `edcff6f`. Both mentions are gone:

- `RELEASING.md` now just says the workflow "runs at 01:00 UTC every Sunday".
- The cron comment in `sunday-release-merge.yml` reads "Early Sunday" and explains the DST drift generically — Actions evaluates cron in UTC with no DST handling, so the local hour it lands on shifts by one across the year, and scheduled runs are queued and delayed anyway.

The schedule itself is unchanged (`0 1 * * 0`).

---

Issue #120, opened 2026-08-31
feat: work in kilograms everywhere, display in a preferred unit from Settings


igorbasko01:
## Problem

`WeightUnit` exists in the model (`lib/features/weight/data/weight.dart`, `typeId: 3`) with `kilograms` and `pounds`, and `Weight.unit` is persisted per entry — but the feature is half-wired:

- **Nothing can produce a pounds entry through the UI.** `AddWeightDialog._submit` (`lib/ui/widgets/add_weight_dialog.dart:44`) hardcodes `'unit': WeightUnit.kilograms`. There is no unit picker and no unit preference anywhere.
- **The only way pounds enter the store is a backup restore.** `WeightBackupCodec.parseWeightCsv` (`weight_backup_codec.dart:67`) parses a unit column and will happily produce `WeightUnit.pounds` rows.
- **Display is inconsistent with itself.** `weight_page.dart:175` hardcodes `'kg'` while `weight_history_tile.dart:45` and `bite_analytics_page.dart:339` render `unit.symbol` from the entry — so on one screen a pounds entry is mislabelled kg, and on another it isn't.
- **There is no conversion anywhere.** No kg↔lbs helper exists, so any computation spanning a mixed-unit history is arithmetic on incompatible numbers.

So the store can hold a unit the UI can't create, can't convert, and renders inconsistently.

## Decision

**Kilograms are the internal unit, everywhere.** Always store kg, always compute in kg. The preferred unit is a **Settings preference** that governs both display and input, converting at the boundary.

**`Weight.unit` is left exactly as it is** — the field stays, the `@HiveField(2)` annotation stays, `WeightUnit` stays. No model change, so:

- no `build_runner` run, no adapter regeneration, no `typeId` risk
- no Hive migration and no new case for `hive_ce_migration_test.dart`
- nothing that currently reads `Weight.unit` stops compiling

The field simply becomes vestigial: always written as `kilograms`, and **never read by any computation**. Treat it as legacy provenance, not as an input.

## Work

**Storage and computation**

- Everything already writes `kilograms`; keep it that way. No writer changes needed.
- Analytics ignore `Weight.unit` entirely and treat every stored value as kg.
[...]

PR #133, opened 2026-09-04
feat: capture the user's height in Settings, in metric or imperial

### LINKED ISSUE #129: feat: capture the user's height in Settings, in metric or imperial
## Problem

The app knows nothing about the user's body beyond a weight series. Any height-derived metric — BMI first among them — is impossible today because there is nowhere to put a height.

Settings also has no preference persistence at all: the only stores are the Hive `weights` box and the Drift bite database, and `shared_preferences` sits in `pubspec.yaml` (line 26) **unused in `lib/`**. So this issue carries the same new plumbing #120 describes; whichever lands first should build it and the other should reuse it.

## Decision

**Centimetres are the internal unit, always.** Store one `double` height in cm; metric/imperial is a display-and-input concern, converted at the boundary. This mirrors #120's "kilograms everywhere, display in a preferred unit" decision rather than inventing a second convention.

**Height has no history.** It's a single current value, not a dated series — no new Hive type, no `typeId`, no Drift table, no migration. A `shared_preferences` key (`height_cm`) is enough.

**One measurement-system preference, not two.** If #120 lands its kg/lbs preference, height should follow the same switch (metric → kg + cm, imperial → lbs + ft/in) rather than letting someone pick pounds and centimetres. If this issue lands first, introduce the preference as a measurement system from the start and let #120 read it.

## Work

**Storage**

- A `height_cm` double in `shared_preferences`, exposed through a small settings store/manager in `lib/features/settings/data/` and provided like the other managers in `lib/main.dart`.
- **Unset is a real state.** No default height — nothing may substitute 170 cm for "not answered". Consumers get `null` and handle it (see the BMI issue's empty state).
- Add an inches↔cm helper in `lib/core/` alongside the kg↔lbs one from #120, with the constant defined once (`1 in = 2.54 cm`, exact).

**Input**

- A Height tile in a new **Profile** (or Body) section on the Settings page, showing the current value in the active system (`178 cm` / `5' 10"`) or a "Not set" affordance, opening a dialog on tap.
- **Metric:** one field in cm, decimals allowed but never required (`178`, `177.5`).
- **Imperial:** two fields — feet and inches — because nobody types their height as a single decimal number of feet. Inches accept `0`–`11` plus a decimal (`10.5`); a value of `12` or more is a validation error rather than a silent roll-over into the next foot.
- **Editing must not drift the stored value.** cm → ft/in → cm round-trips are lossy; leave the stored cm untouched when the fields are unchanged, exactly as #120 requires for weight.
- Validation runs against what was typed, so error messages match the entered unit. Plausible bounds (roughly 50–272 cm) reject typos like a stray decimal point; zero, negative, and unparseable input are errors.

**Backup**

`SerializationService` exports both stores today; a height living only in `shared_preferences` would silently vanish on restore and survive "Clear All Data" — both wrong.

- Add a `profile.csv` entry (`height_cm`) with its own codec in `lib/features/settings/data/`, coordinated by `SerializationService` like the existing three.
- Follow the established optional-entry rule: an archive **without** `profile.csv` leaves the stored height untouched (older backups don't describe it); an archive **with** it replaces the height.
- `clearAllData` clears the height too.

## Notes

- Tests: conversion round-trips at the boundaries; the dialog storing cm unchanged when the field isn't edited; imperial input converting on save; validation rejecting out-of-range and `12`-plus inches; the profile codec for present / absent / empty `profile.csv`; unset height surviving as `null` rather than a default.
- Related: #120 (preferred unit + the same `shared_preferences` plumbing). Blocks the BMI scale issue — that one has nothing to compute without this.

igorbasko01:
## What changed

The app's first preference store, plus a Profile section on Settings that captures a height.

- **`lib/core/units.dart`** — the inch defined once (`2.54 cm`, exact), cm↔in and cm↔ft/in conversions, the `MeasurementSystem` enum, and the display formatters (`178 cm`, `5' 10"`). `roundedFeetInches` is the single place a rounded-up twelfth inch carries into the next foot, shared by the display and the input fields.
- **`lib/features/settings/data/`** — `SettingsRepository` (interface) with `PreferencesSettingsRepository` (`shared_preferences`, the dependency that was in `pubspec.yaml` but unused in `lib/`) and `InMemorySettingsRepository` for tests, plus a `SettingsManager extends ChangeNotifier` that writes through the repository and notifies, like the other managers. Both are provided in `lib/main.dart`.
- **`lib/ui/widgets/height_dialog.dart`** — one cm field in metric, two fields (feet + inches) in imperial, with validation phrased in the unit that was typed.
- **`lib/ui/pages/settings_page.dart`** — a Profile section above Data Management: a Height tile showing the value in the active system or "Not set", and a Metric/Imperial control.
- **`ProfileBackupCodec`** — a `profile.csv` entry (`height_cm`) in the backup zip, coordinated by `SerializationService` alongside the existing three.

## How the issue's requirements are met

**Storage**
- `height_cm` double in `shared_preferences`, behind a repository interface and a manager, provided from `main.dart`.
- Unset is a real state: consumers get `null`; nothing substitutes a default height. The page renders "Not set", and `setHeightCm(null)` removes the key rather than storing a zero.
- Inches↔cm helper in `lib/core/`, with `2.54` written once.

**Input**
- Metric: one cm field, decimals allowed but not required (`178`, `177.5`).
- Imperial: feet and inches as two fields; `12`-or-more inches is a validation error, never a silent roll-over into the next foot.
- Editing doesn't drift the stored value: when no field was touched, the dialog returns the stored centimetres verbatim rather than a value re-derived through a lossy ft/in round trip. Covered by tests in both systems.
- Validation runs against what was typed, so metric says "between 50 and 272 cm" and imperial says `between 1' 8" and 8' 11"`. Zero, negative and unparseable input are errors.

**Backup**
- `profile.csv` codec coordinated by `SerializationService`; an archive without the entry leaves the stored height untouched, an archive with it replaces the height (including restoring "unset" from a backup taken before a height was entered).
- `clearAllData` clears the height too, and `settingsRepo` is a required parameter on every destructive path so a caller cannot skip it.

- **The measurement system is introduced here** (`metric`/`imperial`, one switch), since this issue landed before #120. #120's kg/lbs display preference should read the same `MeasurementSystem` off `SettingsRepository` rather than adding a second preference.
- **A Metric/Imperial control ships with the Height tile.** The issue specifies imperial input but no way to reach it; without a control, imperial would be dead code.
- **The measurement system is not backed up.** `profile.csv` carries only `height_cm`, as the issue specifies — the system is how this device displays figures, not a fact about the user.
- **A height counts as data worth exporting.** Export previously returned "nothing to export" when both stores were empty; a stored height now makes an export non-empty, because unlike the seeded pacing thresholds it is something the user entered.
- **A `ProfileBackup` type rather than a bare `double?`** decoded from the archive, so "the archive says no height" stays distinguishable from "the archive has no profile entry" — the same null-vs-empty contract the bite and pacing codecs rely on.
- No model change, no Hive `typeId`, no Drift table or `schemaVersion` bump — the height has no history.

## Verification

**Flutter is not installed in this environment, so `flutter analyze` and `flutter test` were not run locally** — per `CLAUDE.md`, the toolchain was deliberately not installed and verification is deferred to the `flutter_ci.yml` checks on this PR. The diff was kept small and self-reviewed instead.

Tests added:
- `test/core/units_test.dart` — conversions and round trips at the boundaries, the ft/in split, and the foot-boundary carry that would otherwise read `5' 12"`.
- `test/features/settings/data/settings_manager_test.dart` — unset reads as null, mutations write through and notify, `refresh` picks up a write made behind the manager.
- `test/features/settings/data/preferences_settings_repository_test.dart` — the real `shared_preferences` path: fresh install, round trip, null removing the key, unrecognised stored system falling back to metric.
- `test/features/settings/data/profile_backup_codec_test.dart` — present / absent / empty `profile.csv`, and non-numeric or non-positive values.
- `test/features/settings/data/serialization_service_test.dart` — the height exported, restored, restored-as-unset, left alone by an archive without the entry, and cleared by `clearAllData`.
- `test/ui/widgets/height_dialog_test.dart` — metric and imperial input, the undrifted untouched-field path in both, the foot-boundary pre-fill, 12-plus inches, out-of-range and unparseable input, cancel.
- `test/ui/pages/settings_page_test.dart` — "Not set", the height shown in the active system, the dialog storing what it returns, the system being stored, and a clear leaving no height on the page.

`CLAUDE.md` gains a short note on the new preference store and the fourth backup codec.

A second commit addresses an independent review of this PR: the imperial pre-fill could render `12` inches for a height within a fifth of a millimetre below a foot boundary (fixed, with tests), and `settingsRepo` went from optional to required on the destructive serialization paths.

Closes #129

https://claude.ai/code/session_01HfFWBNz6Axxwc8RdoP7ekY

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,6 +38,8 @@ Key layering for the weight feature:
 
 Dates are treated as day-granular throughout: repository keys and equality normalize to `(year, month, day)`, so "one entry per day" is the invariant. When adding mutation paths, follow the existing pattern (write through the repository, then refresh `_weights` from it).
 
+Settings preferences are a third store, on `shared_preferences` rather than Hive or Drift: `SettingsRepository` (interface, with `PreferencesSettingsRepository` and `InMemorySettingsRepository`) holds the single current values that are not a dated series — the height in centimetres and the `MeasurementSystem` heights are displayed in — and `SettingsManager` exposes them to the UI. Centimetres are the internal unit; imperial is converted at the input/display boundary through `lib/core/units.dart`. An unset height is `null`, never a default.
+
 The bite feature mirrors this shape (interface + manager) on Drift instead of Hive:
 
 - **`BiteDatabase`** (`bite_database.dart`) — two tables: `bites` (append-only `at_ms` epoch-millis log; only the raw timestamp is stored) and `pacing_config` (versioned `b1_s`/`b2_s` thresholds, a slowly-changing dimension seeded with a default on first run).
@@ -46,7 +48,7 @@ The bite feature mirrors this shape (interface + manager) on Drift instead of Hi
 
 ### Backup / restore
 
-`SerializationService` (Settings tab) exports/imports one zip spanning **both** stores. Each dataset owns a per-store codec — `WeightBackupCodec` (`weight.csv`), `BiteBackupCodec` (`bites.csv`, raw `at_ms`), and `PacingConfigBackupCodec` (`pacing_config.csv`, the threshold versions) — and `SerializationService` coordinates them into a single archive. `restoreFromBackup` is the destructive clear-then-restore core, kept separate from file-picker/IO so it stays unit-testable (`clearAllData` is the same clear with nothing restored after it, behind the Settings "Clear All Data" action); it always replaces weights, and replaces bites / pacing config only when the archive actually carries that entry (an older backup missing an entry leaves that store untouched), deduping on import (repeated instants for bites, repeated `effective_ms` for config). Core CSV helpers live in `lib/core/` (`csv_serializer.dart`, `where.dart`).
+`SerializationService` (Settings tab) exports/imports one zip spanning **every** store. Each dataset owns a per-store codec — `WeightBackupCodec` (`weight.csv`), `BiteBackupCodec` (`bites.csv`, raw `at_ms`), `PacingConfigBackupCodec` (`pacing_config.csv`, the threshold versions), and `ProfileBackupCodec` (`profile.csv`, the stored height) — and `SerializationService` coordinates them into a single archive. `restoreFromBackup` is the destructive clear-then-restore core, kept separate from file-picker/IO so it stays unit-testable (`clearAllData` is the same clear with nothing restored after it, behind the Settings "Clear All Data" action); it always replaces weights, and replaces bites / pacing config / profile only when the archive actually carries that entry (an older backup missing an entry leaves that store untouched), deduping on import (repeated instants for bites, repeated `effective_ms` for config). Core CSV helpers live in `lib/core/` (`csv_serializer.dart`, `where.dart`).
 
 ## Testing
 
--- a/lib/core/units.dart
+++ b/lib/core/units.dart
@@ -1,0 +1,55 @@
+/// The system a figure is shown and typed in. Storage stays metric whatever
+/// this says — it is a display-and-input concern, converted at the boundary.
+enum MeasurementSystem { metric, imperial }
+
+/// Exact by definition, and the only place the inch is spelled out.
+const double centimetresPerInch = 2.54;
+
+const int inchesPerFoot = 12;
+
+double inchesToCentimetres(double inches) => inches * centimetresPerInch;
+
+double centimetresToInches(double centimetres) =>
+    centimetres / centimetresPerInch;
+
+/// A length the way a height is said out loud: whole feet plus the inches left
+/// over.
+class FeetInches {
+  final int feet;
+  final double inches;
+
+  const FeetInches(this.feet, this.inches);
+}
+
+FeetInches centimetresToFeetInches(double centimetres) {
+  final totalInches = centimetresToInches(centimetres);
+  final feet = totalInches ~/ inchesPerFoot;
+  return FeetInches(feet, totalInches - feet * inchesPerFoot);
+}
+
+double feetInchesToCentimetres(int feet, double inches) =>
+    inchesToCentimetres(feet * inchesPerFoot + inches);
+
+/// A bare number with at most one decimal and no trailing `.0`, so a whole
+/// height reads `178` rather than `178.0`.
+String formatLengthValue(double value) {
+  final text = value.toStringAsFixed(1);
+  return text.endsWith('.0') ? text.substring(0, text.length - 2) : text;
+}
+
+/// [centimetres] as it reads in [system]: `177.5 cm`, or `5' 10"`.
+String formatHeight(double centimetres, MeasurementSystem system) {
+  if (system == MeasurementSystem.metric) {
+    return '${formatLengthValue(centimetres)} cm';
+  }
+
+  final split = centimetresToFeetInches(centimetres);
+  var feet = split.feet;
+  var inches = split.inches.round();
+  // Rounding 11.6" up would otherwise print 12".
+  if (inches == inchesPerFoot) {
+    feet += 1;
+    inches = 0;
+  }
+  return "$feet' $inches\"";
+}
--- a/lib/core/units.dart
+++ b/lib/core/units.dart
@@ -30,6 +30,17 @@ FeetInches centimetresToFeetInches(double centimetres) {
 double feetInchesToCentimetres(int feet, double inches) =>
     inchesToCentimetres(feet * inchesPerFoot + inches);
 
+/// [centimetres] as feet and inches rounded to [decimals] places, carrying a
+/// rounded-up twelve into the next foot so no display or field ever reads
+/// `5' 12"`.
+FeetInches roundedFeetInches(double centimetres, int decimals) {
+  final split = centimetresToFeetInches(centimetres);
+  final inches = double.parse(split.inches.toStringAsFixed(decimals));
+  return inches >= inchesPerFoot
+      ? FeetInches(split.feet + 1, 0)
+      : FeetInches(split.feet, inches);
+}
+
 /// A bare number with at most one decimal and no trailing `.0`, so a whole
 /// height reads `178` rather than `178.0`.
 String formatLengthValue(double value) {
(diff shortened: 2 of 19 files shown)


igorbasko01:
**Verdict: nothing blocks merge.** The repo's invariants are respected — no Hive `typeId`/`@HiveField` churn, no generated code touched, no Drift table or `schemaVersion` involved (correct: a height is a current value, not a series), weight day-granularity and the write-through-then-notify manager shape are untouched, the new store is injected as `Provider<SettingsRepository>` in `lib/main.dart` and consumed through the interface, and `ProfileBackupCodec` follows the established null-vs-empty optional-entry rule that `BiteBackupCodec`/`PacingConfigBackupCodec` already rely on. Title prefix `feat:` matches a user-visible feature; `pubspec.yaml`/`CHANGELOG.md` were correctly left to release-please. CI (`analyze_and_test`) is green.

Two things worth fixing, neither a merge blocker.

### 1. The imperial dialog can pre-fill `12` inches — `lib/ui/widgets/height_dialog.dart:49-51`

`initState` splits the stored height and writes `formatLengthValue(split.inches)` into the inches field. `formatLengthValue` rounds to one decimal, so a remainder just under 12" is displayed as `12`. `formatHeight` handles exactly this carry (`units.dart:50-54`), but the dialog does not.

Concrete case: store `182.8` cm in metric (the field accepts decimals), switch to Imperial. `182.8 / 2.54 = 71.9685"` → feet 5, inches `11.9685` → the field renders `12`.

- The Height tile reads `6' 0"` while the dialog opens showing `5` ft / `12` in — the two disagree.
- Saving untouched is fine (the `_untouched` short-circuit returns the stored cm), but changing *only* the feet field makes the dialog reject a value the user never typed: `Inches must be less than 12` (`height_dialog.dart:126-128`).

The band is narrow (`[k·30.48 − 0.127, k·30.48)` cm — e.g. 182.76–182.87) but reachable from ordinary metric entry. Applying the same carry as `formatHeight` when seeding the two controllers fixes it, and `test/ui/widgets/height_dialog_test.dart` has no case for a pre-filled imperial height near a foot boundary.

### 2. `settingsRepo` being optional silently drops the profile — `lib/features/settings/data/serialization_service.dart:158`, `:183`

The PR body justifies the optional named parameter on `restoreFromBackup` / `confirmAndRestore` / `clearAllData` as "a required parameter would break call sites and Flutter isn't installed locally, so nobody could catch the compile error before CI". CI is exactly the thing that catches it, and it is green here — so the justification has already expired, while the footgun stays: any future caller that omits `settingsRepo` decodes the archive's height and throws it away, or clears every store *except* the height, with no compile error and no test failure. Both existing call sites already pass it. Making it required now, while there are only two, is cheaper than after the third.

### Smaller notes (no action required)

- Once a height is set there is no way to return it to "not set" from Settings: `_editHeight` (`settings_page.dart:293-305`) stores only a non-null result, and Cancel means "no change". `setHeightCm(null)` is reachable only via Clear All Data or a restore. Given how much the issue and the code invest in unset being a real state, a "Clear" action in the dialog would close the loop — worth a follow-up rather than this PR.
- `int.tryParse` on the feet field (`height_dialog.dart:114`) rejects `5.0` with the generic "Enter your height in feet and inches". Fine, just noting the message is a bit blunt for a plausible typo.
- Test coverage is otherwise good: the codec's present/absent/empty trio, the restore-leaves-it-alone path against a genuinely older weight-only archive, and the undrifted untouched-field path in both systems are all real regression guards rather than shape-checks.

---

igorbasko01:
Both review findings are addressed in 6c9d81d.

**1. Imperial pre-fill could render `12` inches** — confirmed and fixed. Only `formatHeight` carried a rounded-up twelve into the next foot; the dialog's `initState` rounded independently and didn't. Both now go through one `roundedFeetInches(centimetres, decimals)` helper in `lib/core/units.dart`, so a stored 182.8 cm opens as `6` ft / `0` in rather than `5` / `12`. Covered by a unit test on the helper at the foot boundary and a widget test that pre-fills 182.8 cm in imperial, checks the field contents, and edits the feet field through to a save.

**2. Optional `settingsRepo`** — taken. The rationale I gave in the PR description (a required parameter's compile errors couldn't be caught without a local Flutter toolchain) stopped applying the moment CI came back green, and the residual footgun is real: a caller omitting it would silently drop a backup's height or clear every store but the height. It is now `required` on `restoreFromBackup`, `confirmAndRestore` and `clearAllData`, with every call site updated.

The three non-actionable notes are left for the human reviewer: there is no UI path back to "not set" once a height is stored, and the feet field rejects `5.0` via `int.tryParse` with a general message. Both are outside what #129 specifies, and neither is a defect in the paths the issue describes.

---

igorbasko01:
please merge from develop and resolve conflicts

igorbasko01:
Merged in `main` (d04a9d1) — this repo has no `develop` branch, and `main` is both the default branch and this PR's base, so I took that as what you meant. Say the word if you meant something else.

There were no conflicts to resolve: the PR was `behind`, not conflicted. `main` had gained only #132, which adds `.github/workflows/sunday-release-merge.yml` and a `RELEASING.md` section — no overlap with this diff. CI is re-running against the merged head.

---

igorbasko01:
changing between the units of the measuring system are not changing the display of weights, is it something that is going to be done in a different issue?

igorbasko01:
Correct, and yes — that's #120's job, deliberately left out of this PR.

#129 scopes it that way explicitly: *"Units are out of scope here — see #120"* for the weight side, and *"One measurement-system preference, not two… If this issue lands first, introduce the preference as a measurement system from the start and let #120 read it."* So this PR introduces the preference and the switch, and wires it only to the height. Weights stay in kilograms everywhere — the model, the Weight tab, the chart, the history rows, the backup CSV. #120 is the one that adds kg/lbs display, and it should read the same `MeasurementSystem` off `SettingsRepository` rather than adding a second preference; nothing further is needed here to enable that.

The tile is labelled for what it currently does — the subtitle reads "How heights are shown and entered." — so it doesn't promise pounds before #120 lands. If you'd rather it not appear at all until it governs both, I can hide the switch behind the height dialog (imperial input still reachable, no standing control) — say the word. Otherwise this is working as specified.

For the record, both checks are green on `323dbbb`: `analyze_and_test`, and the test-APK `build` you kicked off.

---

PR #134, opened 2026-09-06
feat: weight tab stat tiles show change, not three flavours of lowest weight

### LINKED ISSUE #119: feat: weight tab stat tiles should show change, not three flavours of lowest weight
## Problem

The stat row on the Weight tab (`lib/ui/pages/weight_page.dart`, the `_buildStatCard` row) shows **All Time / 30 Days / 7 Days**, and all three are the same statistic: `getLowestWeight` over a window (`WeightAnalytics.lowestAllTime` / `lowestLast30Days` / `lowestLast7Days`).

Two problems with that:

- It spends the whole row on one measurement. The 7-day lowest in particular just tracks the current weight around and says nothing about direction — it drifts down when you lose and drifts up when you gain, with no indication of which happened.
- The app already computes real change data — `WeightAnalytics.weeklyChanges()` produces 52 weeks of deltas — but only the Home heatmap consumes it. The Weight tab never surfaces a change figure at all.

Meanwhile the current absolute weight isn't on the page anywhere except inside the history list.

## Proposal

Replace the three lowest-weight tiles with three change tiles, and promote the current weight out of the tiles into a hero figure above the chart (so the absolute number becomes *more* prominent than it is today, not less).

| Tile | Value | Sub-label |
|---|---|---|
| **Weekly change** | `-0.6 kg` | `vs. previous week` |
| **30-day trend** | `-0.45 kg/wk` | `≈ -1.9 kg/month` |
| **Vs. low** | `+1.4 kg` | `low 72.3 on 14/03` |

### Weekly change

`mean(most recent complete week) - mean(the week before it)`, over **Sunday-to-Saturday weeks matching the heatmap** — reuse `WeightAnalytics._weekStart` rather than defining a second week boundary, so the Weight tab and the Home heatmap never disagree about which days "this week" covers.

Two deliberate choices here:

- **Means, not endpoints.** Not "today minus the entry 7 days ago" — the store is sparse and day-granular, so an endpoint-to-endpoint delta breaks whenever either end has no weigh-in, which is the common case. Averaging both windows also damps day-to-day water-weight noise. (This differs from `WeeklyWeightChange`, which is intra-week last-minus-first; that one answers "what happened during this week", this one answers "is this week different from last week".)
- **Complete weeks only.** The in-progress week is skipped, so the figure can be up to 6 days stale but is always computed over two full windows — a partial week's mean is being compared against a full week's mean otherwise.

### 30-day trend

Least-squares slope over the last 30 days of entries, expressed as weight per week. Regressing value on day-number handles irregular spacing for free, so gaps in the log don't distort it. The sub-label projects the same slope out to a month.

### Vs. low

`latest - lowestAllTime`, with the date of that low as the sub-label. Reuses the statistic already on the page but reframes it as a distance: near your low it's something to protect, far above it it's a target you've provably hit before.

Note this value is **always ≥ 0** by construction — `lowestAllTime` is computed over all entries including the latest — so it must not use the signed green/red colouring, which would leave it permanently red. Instead:

- **`0` (you are at your all-time low): a celebration icon** — a trophy or similar, shown alongside the value. An icon only, no animation or effect.
- **Any other value: grey.** No red, no green.

### Colour and formatting

The two signed tiles (weekly change, 30-day trend) use the existing convention from `lib/ui/widgets/weight_change_indicator.dart` — down is green, up is `colorScheme.error`, with the matching arrow icon — so they read consistently with the history rows below them. `StatTile` currently hardcodes `colorScheme.primary` for its value, so it needs an optional colour (and icon) override.

Dates go through `lib/core/date_format.dart` so they stay locale-aware — `shortDate` in the tile sub-labels, where space is tight. See #117.

## Empty and degenerate states

**Anything without enough data to state honestly renders `--`.** No partial values, no zeros standing in for absence. That covers:

- **Below the data gate.** Share the gate #115 lands on — replacing `WeeklyWeightChange.minEntries` (an entry count) with a minimum **span**: first-to-last weigh-in at least 3 days apart. Apply the same 3-day span to each of the two week windows *and* to the 30-day regression window. It is a weak gate for a 30-day window — three clustered entries would pass and get extrapolated across the month — but it ships as one rule on purpose; tune it in a follow-up once the tiles have been used for a while.
- **Division by zero in the regression.** If every entry in the window falls on a single day the x-variance is zero and the slope is undefined — guard it and show `--`, never `NaN` or `Infinity`.
- **A stale or missing latest weigh-in.** The hero current-weight figure shows `--` when there is no entry, and follows the same rule rather than presenting an old number as current.
- **No entries at all**, as today.

(The Bite analytics tiles use an em dash `—` for the same state; worth aligning the two eventually, but this issue follows the Weight tab's existing `--`.)

## Notes

- All the computation belongs in `WeightAnalytics` (pure, over the repository) and is exposed through `WeightManager` alongside the existing getters. No model change, no new `typeId`, no Hive migration, no backup-codec impact.
- **Delete `lowestLast30Days` and `lowestLast7Days`** from both `WeightAnalytics` and `WeightManager` once nothing renders them. `lowestAllTime` stays — the Vs. low tile needs it.
- The tiles keep **fixed** 1-week / 30-day windows and deliberately do not follow the History range selector, matching how the current tiles behave.
- Units are out of scope here — see #120, which keeps kilograms as the internal unit and adds a display-only preference. These tiles compute in kg; routing their labels through that preference is #120's job.
- Tests go in `test/features/weight/` against `InMemoryWeightRepository` — no Hive needed. Cover the sparse cases explicitly: missing days inside a window, a window with entries on only one side of the comparison, a window that fails the span gate, all-entries-on-one-day (the zero-variance guard), and a single-entry history.

## Other change stats considered, not in this cut

Parked here so they don't get re-derived later:

- **Total change since first weigh-in** (`latest - first`) — the journey number, trivial to add.
- **Best week** — the biggest weekly drop out of `weeklyChanges`, with its date. Nearly free, since those 52 weeks are already computed for the heatmap.
- **Volatility** — `max - min` over 30 days. Reads as dull on its own but it's the calibration for everything above: if the scale swings 1.2 kg, a 0.3 kg weekly change is noise. Probably better as a sub-label than its own tile.
- **Consistency** — days logged out of the last 30, or the current streak. Not a change stat, but it's the honest caveat on all of them: a `-0.8 kg` derived from 3 entries in a month deserves an asterisk.

igorbasko01:
Replaces the Weight tab's All Time / 30 Days / 7 Days lowest-weight tiles with three change tiles, and promotes the current weight out of the tiles into a hero figure above the chart.

## What changed

**`WeightAnalytics`** (pure, over the repository) gains:

- `weeklyChange({asOf})` — `mean(most recent complete week) − mean(the week before it)`, over the same Sunday-to-Saturday weeks `weeklyChanges()` uses (it reuses `_weekStart`, so the tab and the Home heatmap can never disagree about which days a week covers). Means, not endpoints; the in-progress week is skipped.
- `trendPerWeek({asOf})` — least-squares slope of value on day number over the last 30 days, expressed as kg/week. Regressing on day number rather than list position means gaps in the log stretch the x axis instead of distorting the slope. Guards a zero x-variance denominator.
- `latestEntry`, `lowestEntry`, `changeFromLowest` — the hero figure, the date of the low, and the distance to it. All three ignore weigh-ins dated ahead of today, matching the week and 30-day windows.
- `minSpanDays` — aliased to `WeeklyWeightChange.minSpanDays`, so every window here gates on the same span the Home heatmap does.

`lowestLast30Days` and `lowestLast7Days` are deleted from both `WeightAnalytics` and `WeightManager`. `lowestAllTime` stays, as the issue asks, but now derives from `lowestEntry` rather than running a second independent search for the same figure.

**`StatTile`** gains optional `valueColor` and `icon`, and draws its value inside a `FittedBox(scaleDown)` so a long signed reading like `-0.45 kg/wk` stays on one line instead of wrapping and breaking the row's shared height.

**`WeightPage`** renders the hero figure, the three tiles, and the `--` fallbacks.

| Tile | Value | Sub-label |
|---|---|---|
| Weekly change | `-0.6 kg` | `vs. previous week` |
| 30-day trend | `-0.45 kg/wk` | `≈ -1.9 kg/month` |
| Vs. low | `+1.4 kg` | `low 72.3 on 14/03` |

## Meeting the issue's requirements

- **Weekly change** — means of two complete Sunday-to-Saturday windows, sharing `_weekStart` with the heatmap.
- **30-day trend** — least-squares slope on day number, per week, with the month projection in the sub-label.
- **Vs. low** — `latest − lowest`, dated. Always ≥ 0, so it deliberately does **not** use the signed colouring: it is grey at any value, and gets a trophy icon (`Icons.emoji_events`, no animation) at exactly `0`.
- **Colour and formatting** — the two signed tiles follow `weight_change_indicator.dart`: down is green, up is `colorScheme.error`, with the matching arrow; a flat reading uses `colorScheme.outline` and the remove icon. Dates go through `core/date_format.dart` (`shortDate` in the tile sub-labels, `fullDateWithWeekday` under the hero).
- **The shared data gate** — `WeightAnalytics.minSpanDays` is `WeeklyWeightChange.minSpanDays`, the span gate #115 landed in #131. One constant, one rule, applied to each of the two week windows and to the 30-day regression, exactly as the issue asks.
- **Empty and degenerate states** — everything without enough data renders `--`: below the span gate, a zero-variance regression, and no entries at all. The hero shows `--` and "No weigh-ins yet" with no entries.
- **No model change** — no `typeId`, no Hive migration, no `build_runner`, no backup-codec impact.
- **Fixed windows** — the tiles ignore the History range selector, as the old ones did.
- **Units out of scope** — everything computes and renders in kg, for #120 to route later.

- **The span gate.** This PR was opened while #131 was still in review, so it originally defined its own 3-day constant rather than colliding with the rewrite of `WeeklyWeightChange` — and this description carried a note about the two surfaces disagreeing in the meantime. #131 has since merged, main is merged in here, and `WeightAnalytics.minSpanDays` now simply points at `WeeklyWeightChange.minSpanDays`. Same threshold, same UTC-midnight day counting, so nothing moved behaviourally; that earlier divergence no longer exists.
- **The hero is dated rather than staleness-gated.** The issue says the hero "follows the same rule rather than presenting an old number as current", but names no staleness cutoff. Inventing one would be scope the issue doesn't specify, so the hero instead carries `as of <date>` under the figure: an old reading is visibly old without a threshold anyone has to agree on.
- **Future-dated weigh-ins are excluded** from the hero and the low, matching the week and 30-day windows. The date picker reaches years ahead, so without this a weigh-in logged for next week became the tab's "current" weight.
- **The trend's month projection is `perWeek × 30 / 7`**, so the sub-label is the same slope the value states rather than a second regression.
- **Displayed values are rounded before their sign, colour, and arrow are chosen**, so a −0.04 kg week reads as flat rather than printing `-0.0` next to a green down arrow.
- **The low is dated at the *last* time it was reached** when the same value appears more than once, so the trophy case shows today rather than a date from months ago.
- **`FittedBox` on `StatTile`'s value** — needed for the longer readings here; it only scales when the text would otherwise not fit, so the Bite tab's existing tiles are unaffected.

## Verification

Flutter is not installed in this environment, and per `CLAUDE.md` it was not installed — verification is deferred to the `flutter_ci.yml` checks on this PR (`flutter analyze` + `flutter test`).

Tests added in `test/features/weight/weight_analytics_test.dart` against `InMemoryWeightRepository`, covering the sparse cases the issue calls out: missing days inside a window, entries on only one side of the comparison, a window failing the span gate, exactly-three-days-apart clearing it, a single weigh-in (the zero-variance case), the in-progress week being ignored, irregular spacing, the 30-day window's boundaries, and future-dated entries being left out. `test/ui/pages/weight_page_test.dart` covers the rendered tiles in both directions, the trophy state, an older low, the gate fallback, and the empty state. `weight_manager_test.dart` swaps its deleted-getter tests for the new ones.

Closes #119

https://claude.ai/code/session_01Wr3rY8oUk5aQjoyj35chRZ


igorbasko01:
**Verdict: nothing here blocks merge.** No Hive `typeId` or `@HiveField` touched, no generated file hand-edited, no Drift schema change, no backup-codec impact, no manual `pubspec.yaml`/`CHANGELOG.md` bump, and `feat:` is the right prefix for a user-visible tab change. Day-granularity and the repository-interface layering are respected — `WeightAnalytics` reads only through `WeightRepository`, and the new tests use `InMemoryWeightRepository`. The DST care in `_daysBetween` (UTC component subtraction) and `_shiftWeeks` (calendar-component stepping) is correct, and `_roundTo` before choosing sign/colour/arrow genuinely does prevent a `-0.0` next to a green down arrow.

Four things worth looking at, most serious first.

---

### 1. The Weight tab and the Home heatmap *can* disagree about a week — and the PR body says they can't

`lib/features/weight/data/weight_analytics.dart:14` (`minSpanDays = 3`) vs `lib/features/weight/data/weekly_weight_change.dart:12` (`minEntries = 4`).

The body says:

> `WeightAnalytics.minSpanDays` uses the same 3-day rule; once #115 merges, the two constants should collapse into one — a one-line follow-up, and no behaviour changes when they do.

`WeeklyWeightChange.minEntries` is on `main` today and it is **4 weigh-ins**, not a 3-day span. These are different rules over the same Sunday-to-Saturday weeks, so:

- A week with weigh-ins on Sun / Wed / Sat — 3 entries, 6-day span — is **blank** on the Home heatmap (`3 < minEntries`) but is **counted** by `_gatedMean` for the Weekly change tile. So Home can show the last two weeks as "no data" while the Weight tab prints a confident `-0.6 kg / vs. previous week` computed from exactly those two weeks.
- Conversely a week with Mon/Tue/Wed/Thu — 4 entries, 3-day span — clears both, and Mon/Tue/Wed — 3 entries, 2-day span — clears neither.

The sharing claim holds for *which days a week covers* (`_weekStart` really is shared), but not for *whether a week has enough data*. Worth either fixing the body's note or, better, making the follow-up explicit that collapsing the two constants **will** change one surface's behaviour — otherwise whoever picks up that "one-line follow-up" will silently move the heatmap's gate from 4 entries to a 3-day span.

### 2. Future-dated entries drive the hero and Vs. low, but are excluded from the other two tiles

`weight_analytics.dart:23` (`latestEntry`) and `:30` (`lowestEntry`) reduce over the unfiltered `getAllWeights()`.

`AddWeightDialog` sets `lastDate: DateTime(2100)` (`lib/ui/widgets/add_weight_dialog.dart:67`), so a future date is reachable with a mis-tap on the year. The rest of the file already treats that as a known hazard — `trendPerWeek` bounds its window at `today + 1`, and `weeklyChanges` has an explicit `if (start.isAfter(weekStarts.last)) continue` for exactly this. The new getters don't.

Concrete case: a user logs `2027-03-14, 95.0` by accident. The hero reads `95.0 kg / as of Sun, 3/14/2027` forever, `Vs. low` measures the gap from that phantom entry, while `Weekly change` and `30-day trend` correctly ignore it — so three tiles disagree with the big number above them. Clamping `latestEntry` to `<= today` (and, arguably, `lowestEntry` too) would make the whole tab consistent.

### 3. `lowestAllTime` is now dead, and the body's reason for keeping it is wrong

> `lowestAllTime` stays, since the Vs. low tile needs it.

The Vs. low tile uses `changeFromLowest` and `lowestEntry` (`lib/ui/pages/weight_page.dart:55-58`); it never reads `lowestAllTime`. After this PR, `WeightAnalytics.lowestAllTime` and `WeightManager.lowestAllTime` are referenced only by `weight_manager_test.dart`. Not harmful, but it leaves two independent paths computing the same number — the cached `getLowestWeight()` in `WeightRepositoryHelper` and the uncached reduce in `lowestEntry`. Either drop `lowestAllTime` with the other two getters, or express it as `lowestEntry?.value` so there is one source.

### 4. Nits

- **`trendPerWeek`'s `variance == 0` guard is unreachable** (`weight_analytics.dart:112`). `_spansEnough` already requires a ≥3-day span, which means ≥2 distinct days, which means non-zero x-variance. The body attributes the "single weigh-in" test to this guard, but that test passes through the span gate. Harmless defensive code — just not the thing the body says it is.
- **Unit hardcoding, pre-existing.** `_CurrentWeight` prints `'${current.value.toStringAsFixed(1)} kg'` while `Weight.unit` can be `pounds` (the dialog returns `result['unit']` and `addWeight` stores it verbatim). The history list already does this, so the PR isn't introducing the bug — but it does put it in the tab's largest text. Fine to leave for #120 as the body says; flagging so it's on the list.
- **Rebuild cost.** `WeightPage.build` now hits `getAllWeights()` five times (`latestEntry`, `lowestEntry`, and `changeFromLowest` which calls both again) plus two range scans, on every notify. Irrelevant at realistic weigh-in volumes; mentioning only because `changeFromLowest` recomputing both getters is free to avoid.

---

The test coverage is good and hits the sparse cases the issue asked for. One note on `rampedManager`: the 28-day ramp is enough for the worst weekday (today = Saturday needs back to `today-20`), so it isn't weekday-flaky. `flutter analyze`/`flutter test` were still in progress at review time and Flutter isn't installed here, so correctness of the numeric assertions is on CI.

---

igorbasko01:
Thanks — acting on the review. Three of the four findings were right; here's what changed and what didn't.

**Fixed and pushed (b966267):**

- **Future-dated entries (finding 2).** Correct, and reachable through the date picker. `latestEntry`, `lowestEntry` and `changeFromLowest` now read through a `_loggedEntries` filter that drops anything dated ahead of today, so all four surfaces measure the same set. Covered by two new tests.
- **Two paths to the all-time low (finding 3).** Also correct — the tile uses `lowestEntry`, so `lowestAllTime`'s call to `getLowestWeight()` was a second, independent search. It now derives from `lowestEntry`. The getter itself stays, since the issue asks for it to; the PR body no longer claims the tile needs it.

**Corrected in the description, not the code (finding 1).** The important half of this is right and my description was wrong: `WeeklyWeightChange.minEntries = 4` on `main` is an entry count, not a 3-day span, so the heatmap and these tiles really can disagree — a Sun/Wed/Sat week is blank in one and counted in the other. What the body said about the follow-up being a behaviour-preserving no-op was simply incorrect, and it's now rewritten to state the divergence and its consequence outright.

I'm still not changing `WeeklyWeightChange` here. #131 (for #115) rewrites that class wholesale, including introducing the span gate; replacing `minEntries` in this PR would collide with it directly. The single gate the issue wants arrives when #131 lands, and it's a real change to heatmap behaviour that belongs in whichever PR owns that class.

**Not changed (finding 4).** The `variance == 0` guard is unreachable behind `_spansEnough` today, but the issue asks explicitly for a guard against `NaN`/`Infinity`, and it's one line that stays correct if the gate is ever tuned. The hardcoded `kg` in the hero matches every other display site on the tab and is #120's to route.

---

igorbasko01:
please merge from main and resolve conflicts

igorbasko01:
Done — `main` is merged in as of d2ef814, and the PR is no longer conflicted.

The only conflict was in `weight_analytics.dart`: #131 removed `_change` in favour of `WeeklyWeightChange` deriving everything from the week's entries, so I took that removal and kept this branch's window helpers.

Worth knowing, since it closes the one open question this PR was carrying: **#131 landing means the span gate is now genuinely shared.** `WeightAnalytics.minSpanDays` is a plain alias for `WeeklyWeightChange.minSpanDays` rather than a 3-day constant of its own — same threshold, same UTC-midnight day counting — so the "the tiles and the heatmap can disagree" caveat that used to be in the description is gone, and #119's ask for one gate across both surfaces is actually met. The description is updated to match.

CI is re-running on the merge commit now.

---

PR #135, opened 2026-09-07
feat: work in kilograms everywhere, display in a preferred unit from Settings

### LINKED ISSUE #120: feat: work in kilograms everywhere, display in a preferred unit from Settings

igorbasko01:
Kilograms become the internal unit throughout — stored, computed, and passed between widgets. The unit a weight is *shown and typed in* is a new Settings preference, converted at the display and input boundary and nowhere else.

## What changed

**Storage and computation**
- `Weight.unit` is left exactly as it was: same field, same `@HiveField(2)`, same `typeId: 3`. No `build_runner` run, no adapter regeneration, no Hive migration, no new case in `hive_ce_migration_test.dart`. It is now legacy provenance that no computation reads.
- `WeeklyWeightChange.level` no longer doubles its bucket bounds for pounds — with kg-only data that branch was unreachable. `WeeklyWeightChange.unit` stays in place.

**Settings preference (new plumbing, not a new dependency)**
- `SettingsRepository` (interface) / `PreferencesSettingsRepository` (production, on the already-declared but previously unused `shared_preferences`) / `InMemorySettingsRepository` (tests), surfaced through `SettingsManager extends ChangeNotifier` and wired in `lib/main.dart`.
- Reads are synchronous — `open()` loads the stored value once — so a widget renders the preference without an await.
- Settings gains a "Units" section with a kg / lbs segmented picker.

**Conversion**
- `lib/core/weight_units.dart` defines `1 lb = 0.45359237 kg` once, plus the two pure conversions.
- A `WeightUnit` extension in `weight.dart` (`fromKilograms` / `toKilograms` / `format`) is the single step between the store and the screen.

**Input** — `AddWeightDialog` takes the preferred unit, labels the field with it, and converts to kilograms on save, so a pounds user types `160` and the store gets `72.5748`. Decimals are accepted but not required (`160` and `159.8` both work). Validation runs on the number as typed.

**Display sites** — every place a weight reaches the screen now goes through the preference rather than `Weight.unit`: the Weight-tab stat tiles and history rows, `weight_history_tile`, `weight_change_indicator`, `weight_chart` (axis + tooltip), `daily_bites_chart` (right axis, tooltip, legend, semantics), `bite_analytics_page`'s weight line, and the Home heatmap readout (`weekly_change_summary`).

**Backup format** — export writes `date,weight_kg` and drops the unit column, so the file is self-describing. Import stays as permissive as before and reads every value as kilograms; any `unit` column in an older archive is ignored rather than honoured.

## Acceptance criteria

| From the issue | How it is met |
| --- | --- |
| Kilograms internal, everywhere | Analytics, deltas, chart plotting and widget parameters are all kg; `Weight.unit` is read nowhere |
| `Weight.unit` untouched | Field, annotation and enum unchanged; `weight.g.dart` not regenerated |
| Preference governs display *and* input | `SettingsManager.weightUnit` feeds both the display sites and `AddWeightDialog` |
| kg↔lbs helper in `lib/core/`, constant defined once | `lib/core/weight_units.dart` |
| Input converts on save | `AddWeightDialog._submit` pops kilograms |
| Decimals allowed but not required | `160` → 72.5748 kg, `159.8` → 72.4841 kg, both covered by tests |
| Editing must not drift the stored value | An untouched field returns the stored kg verbatim instead of round-tripping through the one-decimal display; test asserts byte-identical `72.57432` |
| Validation runs against the entered unit | Parsed and range-checked before conversion |
| Every display site goes through the preference | See "Display sites" above |
| Export `date,weight_kg`, no unit column | `WeightBackupCodec.generateWeightCsv`, asserted in tests |
| Import permissive, everything read as kg | Legacy `value` column still read, `unit` ignored; a `pounds` row restores as that same number in kg |
| `feat:`, not breaking | Persisted model untouched; every existing archive still restores |

- **The preference reuses the existing `WeightUnit` enum** rather than introducing a parallel display-unit type. It is the smallest option and keeps one vocabulary for "kg or lbs".
- **Display widgets take the unit as a parameter; pages read it from the provider.** This matches `WeightChangeIndicator`, which already took a `WeightUnit`, and keeps the widgets testable without a provider. The two private widgets inside `bite_analytics_page.dart` read it from context instead, since threading it through their constructors bought nothing.
- **Two display sites the issue did not enumerate are included**, under its own rule that every weight on screen goes through the preference: the Home heatmap readout (`weekly_change_summary`, which was reading `week.unit`) and the `daily_bites_chart` legend (`Weight (kg)`).
- **`weeklyChangeSummary`'s `locale` moved from positional to named** so the new `unit` argument could join it without an awkward parameter order.
- **The pounds bucket-scaling in `WeeklyWeightChange.level` is removed** rather than left dead, as the issue permits.
- **Settings persistence is new plumbing here.** Note that #133 (for #129) independently introduces a `SettingsRepository`/`SettingsManager` of the same shape for height — whichever lands second will need those two files merged into one. Named identically here to make that a union rather than a rename.

## Verification

Flutter is not installed in this environment and `CLAUDE.md` says not to install it, so `flutter analyze` and `flutter test` were **not** run locally — verification is deferred to the `flutter_ci.yml` checks on this PR. The diff was self-reviewed instead, and the conversion arithmetic behind each new expectation was computed independently rather than read off an implementation.

Tests added or updated: `weight_units_test` (constant, both directions, round-trip, formatting), `settings_manager_test`, `preferences_settings_repository_test` (unset default, survives a reopen, unknown stored unit), `weight_backup_codec_test` (new column shape, legacy `value` column, a legacy `pounds` row restoring unchanged), `add_weight_dialog_test` (unit label, prefill in the preferred unit, whole-number and one-decimal pounds input, the no-drift untouched edit, a deliberate edit), `settings_page_test` (the picker starts on the stored unit and writes through), plus pounds-preference display assertions on the Weight page, Home history rows, `daily_bites_chart`, `weekly_change_summary` and `weight_change_indicator`.

Closes #120

https://claude.ai/code/session_01JevhE2XRTsBHr2jgaPtWKu

---
_Generated by [Claude Code](https://claude.ai/code/session_01JevhE2XRTsBHr2jgaPtWKu)_


igorbasko01:
**Verdict: nothing in the code blocks merge — but do not merge this and #133 without deciding which preference model wins first.** CI is green, the Hive model and its generated adapter are correctly untouched, no Drift or schema surface is involved, day-granularity and the repository-then-notify mutation shape are unchanged, and `pubspec.yaml`/`CHANGELOG.md` are left to release-please. The conversion itself is clean: every display site now goes through `SettingsManager`, and the one sign-sensitive consumer (`WeightChangeIndicator`) converts by a positive factor so its gain/loss branches are unaffected.

Findings, most serious first.

---

### 1. This PR and #133 both create the same four files, with different designs

`lib/features/settings/data/{settings_repository,preferences_settings_repository,in_memory_settings_repository,settings_manager}.dart` are created by **both** this PR and the open #133, both branched off `5fcd67b`. So are edits to `lib/main.dart`, `lib/ui/pages/settings_page.dart`, `test/ui/pages/settings_page_test.dart`, `test/features/settings/data/settings_manager_test.dart` and `test/features/settings/data/preferences_settings_repository_test.dart`. Whichever merges second conflicts on all of them.

The PR body flags the file overlap, but the divergence is larger than a merge: #133 introduces a `MeasurementSystem` (metric/imperial) enum and states in its own body that *"#120's kg/lbs display preference should read the same `MeasurementSystem` off `SettingsRepository` rather than adding a second preference."* This PR instead adds an independent `weight_unit` preference over the existing `WeightUnit` enum.

Merge both as-is and Settings ships two unrelated unit switches — a Metric/Imperial control governing height and a Kilograms/Pounds control governing weight — that can disagree with each other. That is a product decision, not a conflict resolution, so it is worth settling before either lands rather than in whoever's rebase comes second.

### 2. `weight.csv` written by this build wipes the weight store when restored by an older build

`generateWeightCsv` now emits `date,weight_kg` and drops `value` entirely (`lib/features/settings/data/weight_backup_codec.dart:51-57`). Reading is tolerant in the old→new direction (`item[valueColumn] ?? item[legacyValueColumn]`, line 65), but the new→old direction is not:

- every released build's `parseWeightCsv` reads only `item['value']`, gets `null` for every row, and `continue`s;
- `SerializationService.restoreFromBackup` clears unconditionally and then restores whatever parsed — `weightRepo.clear()` at `serialization_service.dart:192`, followed by a loop over an empty list.

Concrete case: a user on this build exports a backup, reinstalls or sideloads an earlier APK (`build_install_release.sh` makes that a normal flow here), and restores. The archive is valid, `weight.csv` is present, and the restore reports success — with zero weights left and the previous contents already cleared. Bites and pacing config survive, so it reads as a partial restore rather than a corrupt file.

Cheap fix if you want the safety: keep writing the legacy `value` column alongside `weight_kg` for a release or two. Both are kilograms now, so the duplicate column costs nothing but bytes, and old builds keep restoring correctly.

### 3. A preference that fails to persist reverts silently

`PreferencesSettingsRepository.setWeightUnit` assigns `_weightUnit = unit` *before* awaiting `_preferences.setString` (`preferences_settings_repository.dart:40-44`), and the picker calls it without awaiting or catching (`settings_page.dart:274-275`). If the platform write throws, the in-memory value and the UI have already switched, nothing surfaces the failure, and the app comes back on the old unit at next launch. Awaiting the write before mutating `_weightUnit` makes the two agree in every case.

### 4. `Provider<SettingsRepository>` is registered but never read

`lib/main.dart:45` exposes the repository, and nothing resolves it — every consumer goes through `SettingsManager`. `WeightRepository` and `BiteRepository` are provided because the Settings page reads them directly; this one has no such caller. Harmless, but it is wiring that has to be maintained and merged (see #1) for no consumer.

### 5. Nit: pounds axis ticks land on unrounded numbers

`WeightChart` leaves `interval` unset, so fl_chart picks tick positions from the kilogram min/max and only the label is converted (`weight_chart.dart:69`); `DailyBitesChart`'s right axis does the same with `interval: maxY / 4` (`daily_bites_chart.dart:193, 200-202`). A pounds user therefore gets ticks like `158.7 / 163.1 / 167.6` rather than round values. Cosmetic, and only if it bothers you.

---

Two things I checked and am satisfied with, since they are the usual traps here:

- **The `_prefilled` guard in `AddWeightDialog`** (`add_weight_dialog.dart:30-43, 51-60`) is the right call and correctly scoped: it only short-circuits when `initialWeight != null`, so a fresh entry with an empty field still falls through to parsing and validation, and a date-only edit returns the stored kilograms byte-for-byte. It also fixes a pre-existing kg-side drift, where an untouched `75.55` used to be saved back as `75.5`.
- **Legacy `pounds` rows now import as kilograms** — a documented decision in the PR body, and no reachable commit ever wrote a pounds weigh-in (the dialog has hard-coded `WeightUnit.kilograms` for as long as this repo's history goes back), so the exposure is theoretical. Worth knowing that `test/features/weight/hive_ce_migration_test.dart` still asserts a `pounds` row round-trips through the box — correctly left alone as a compatibility contract, but it now guards a state nothing in the app can read as intended.

Title `feat:` is right for the change.

---

igorbasko01:
Acted on the review above. Two findings were defects in code this PR introduces and are fixed in `54631cc`:

- **`setWeightUnit` mutated memory before awaiting the write** — now writes the store first, so a failed write leaves the preference unchanged rather than serving a value that is gone by the next launch.
- **`Provider<SettingsRepository>` was registered but never resolved** — removed; `SettingsManager` is built from the instance directly.

The other three I've deliberately left as-is, for you to weigh:

- **Collision with #133.** Real, and called out in the PR description. Both branches create the same four `settings/data` files, and #133's `MeasurementSystem` covers the same ground as this PR's `WeightUnit` preference. Resolving it means building one on top of the other, which isn't something this branch can do while #133 is unmerged — whichever lands second should read the other's preference rather than keep two switches. Your call which order.
- **Backup forward-compatibility.** The reviewer is right that this is sharper than #120's own framing: a `weight_kg` archive restored on an older build doesn't just skip rows, it wipes weights, because `restoreFromBackup` clears before restoring an empty parse. #120's Compatibility section knowingly accepted the forward-compat edge, so I've implemented the format it specified rather than overriding that decision — but the severity is worth a second look, and emitting both `weight_kg` and `value` for a release or two would close it cheaply if you want it.
- **Irregular pounds axis ticks.** Real but cosmetic, and fixing it means choosing chart intervals in the display unit — chart logic this PR otherwise doesn't touch.

---

igorbasko01:
please merge from main and resolve conflicts

igorbasko01:
Merged `main` in and resolved (`4d692a0`). #129 and #134 both landed first, so this rebuilds onto them rather than beside them — and #129 settles the collision the review flagged:

> **One measurement-system preference, not two.** … If this issue lands first, introduce the preference as a measurement system from the start and let #120 read it.

It did land first, so **this branch's parallel kg/lbs preference is gone entirely**: the four `settings/data` files, the second Settings picker, and `lib/core/weight_units.dart` all give way to main's `SettingsRepository` / `MeasurementSystem` / `lib/core/units.dart`. Metric shows kilograms, imperial pounds, and nobody can pick pounds alongside centimetres. The kg↔lb constant and conversions moved into `lib/core/units.dart` beside the inch ones as a `MeasurementSystem` extension; every display site that took a `WeightUnit` now takes the system.

**#134's stat tiles are re-plumbed, not reverted.** Weekly change, 30-day trend and vs-low each convert *before* rounding, so the digits, sign, colour and arrow still agree with one another in either system — a `-0.7 kg` week reads `-1.5 lbs`, and the trend's `-0.70 kg/wk ≈ -3.0 kg/month` becomes `-1.54 lbs/wk ≈ -6.6 lbs/month`.

Settings' Units row already existed for height, so only its subtitle changed — it now says it governs weights too.

Net effect: the PR shrank from 36 files / +878 to 26 / +652, since the duplicated plumbing is no longer being added. Two of the review's five findings are now resolved by construction (the #133 collision, and the `Provider<SettingsRepository>` I'd added). Still open for you: the backup forward-compatibility severity, and the irregular imperial axis ticks.

---

lib/features/settings/data/weight_backup_codec.dart
```diff
@@ -9,6 +8,15 @@ class WeightBackupCodec {
...
   static const String weightFileName = 'weight.csv';
 
+  /// The value column, named for its unit so the file says what its numbers
+  /// mean.
+  static const String valueColumn = 'weight_kg';
+
+  /// The value column older archives used, alongside a `unit` column. Its
+  /// values are taken as kilograms whatever that unit said — the app itself
+  /// only ever wrote kilograms.
+  static const String legacyValueColumn = 'value';
```
igorbasko01: It always been kgs in the file. I think there is no need for something special for old values. Wdyt?

igorbasko01:
I'm not sure I understand what is going on. Didn't we plan that we always store the weight in kg, and just display the using the user's preferred unit?

lib/features/settings/data/weight_backup_codec.dart
igorbasko01: Agreed, and done in `edca0b4` — though I kept the column name rather than deleting the fallback, because those two are the same fix from different ends.

You're right on the facts: I checked back through the history and the column has been `value` in kilograms in every release, so nothing needs reinterpreting on import. The only reason the fallback existed was that I'd *renamed* the column to `weight_kg`. Deleting the fallback while keeping the rename is the one combination that actually breaks: an older build finds no `weight_kg` column, parses zero rows, and `restoreFromBackup` has already cleared by then — so restoring a new archive on an older APK would wipe the weight history rather than just fail.

Keeping the name is the simpler half of that trade. There's no legacy path to read, and the archive moves in both directions.

The `unit` column still goes, for exactly your reason — it was always `kilograms`. Dropping it is safe in both directions too: an older build already defaults a missing unit to kilograms (there's a test on `main` for that), and a new build reading an archive that still carries the column keeps the number and ignores what it says.

Net effect on the codec is now just this, against `main`:

```diff
-              'unit': w.unit.name,
...
-      final unitStr = item['unit'] as String?;
...
-        WeightUnit unit = WeightUnit.kilograms;
-        if (unitStr != null) { ... }
-        weights.add(Weight(date: date, value: value, unit: unit));
+        weights.add(Weight(date: date, value: value));
```

This also closes the forward-compatibility finding the reviewer raised further up the PR, so nothing is left outstanding there.

---

igorbasko01:
Yes — that's exactly what this does, and it hasn't changed. Sorry for the noise; I buried the plan under merge chatter.

**The plan, as shipped here:** every weight is stored and computed in kilograms. The only conversion is at the display and input boundary. `Weight.unit` stays on the model exactly as it was — same field, same `typeId` — but nothing reads it any more.

**The one thing that did change** is *which* preference picks the display unit. I'd originally added a new kg/lbs preference. Then #129 merged first, and it had already settled this:

> **One measurement-system preference, not two.** … If this issue lands first, introduce the preference as a measurement system from the start and let #120 read it.

So I dropped my preference and this now reads the `MeasurementSystem` you already have in Settings: metric → kg, imperial → lbs. That's why there's no second picker — the existing Units row governs weights as well as heights now.

**The backup thread is a small side matter**, not a change of plan. The CSV has always held kilograms. I had briefly renamed its column `value` → `weight_kg`; you pushed back, and you were right, so I reverted that. The only backup change left is dropping the vestigial `unit` column, which was always written as `kilograms`.

CI is running on `edca0b4`; everything before it was green.

---

lib/features/settings/data/weight_backup_codec.dart
igorbasko01: what about this one?

igorbasko01: This thread is marked outdated — the lines it was anchored to (`weight_kg` / `legacyValueColumn`) are gone as of `edca0b4`, so GitHub is showing you code that no longer exists. That's probably what looks unresolved.

The whole file is now 69 lines, and the only difference from `main` is that `unit` is no longer written or read:

```dart
String generateWeightCsv(List<Weight> weights) {
  final items = weights
      .map((w) => {
            'date': w.date.toIso8601String(),
            'value': w.value,
          })
      .toList();
  return CsvSerializer.toCSV(items);
}
```

No special handling for old values, and no column rename. CI is green on it.

If you meant something else by "this one", point me at the line or paste it and I'll take a look.

---

igorbasko01:
Is passing around this measurement system is a good approach? Maybe only the weight manager should have a reference to the measurement system, and the weight should return the relevant unit and value, instead of each page referencing the measurement system?

igorbasko01:
Fair question. Short version: I'd keep it as-is here, but not because the current shape is sacred — because `WeightManager` specifically can't be the single owner, and because this pattern is already yours.

**It's the convention #129 set.** `lib/ui/widgets/height_dialog.dart` on `main` already takes a `MeasurementSystem` parameter and converts at the boundary. This PR followed that rather than inventing a second style.

**Three things block moving it into `WeightManager`:**

1. **The bite side never touches `WeightManager`.** `bite_analytics_page` gets its `Weight` objects from `WeightRepository` via `BiteAnalyticsController`. Its weight line and the chart's weight overlay would still need the system from somewhere else — so you'd get both patterns instead of one.

2. **`MeasurementSystem` isn't a weight preference.** It governs height too. Putting it in `WeightManager` makes that class the owner of something half of which it has nothing to do with, and height would keep reading `SettingsManager` anyway.

3. **Pre-converted values break the charts.** They deliberately plot in kilograms and convert only the labels, because their constants are kg-denominated: `weight_chart` pads the y-axis ±2, `daily_bites_chart` ±1, and `WeeklyWeightChange.level` buckets at 0.25 / 0.5 / 1.0. Hand those widgets pounds and every one of those numbers silently means something else. Same reason the stat tiles convert *before* rounding rather than after.

   And if the manager returned a formatted string instead of a number, it'd be doing presentation — while the charts would still need the raw value to plot.

**That said, your underlying complaint is real** — ten widgets taking a `system` parameter is a lot of threading. The version of your idea that does work is to push the *read* down instead of the value: let the display widgets do `context.watch<SettingsManager>().measurementSystem` themselves, the way `_WeightLine` and the bite page already do. No page has to remember to pass it, kilograms stay the data everywhere, and the charts keep their raw numbers. The cost is that those widgets stop being pure — every widget test needs a provider around it.

My suggestion: leave this PR as it is (it's green, and it matches height), and if the threading bothers you, do the context-read conversion as its own change covering weight *and* height together, so the two stay consistent. Widening this PR to do half of it would leave the codebase in a worse spot than either end.

Happy to do it either way — say the word and I'll push it here, or open it as a follow-up.

---

igorbasko01:
I was thinking that maybe the weight should have some extension function that asks for the measurementSystem when requesting the system unit or something.

igorbasko01:
That'd read nicely where it applies — but it's a readability choice, not a structural one, and I want to flag that it doesn't fix the thing you raised first.

**It doesn't reduce the threading.** `weight.valueIn(system)` needs `system` in scope exactly as much as `system.formatWeight(weight.value)` does. It swaps the receiver, not who has to know the preference.

**It covers about a third of the sites.** There are 20 conversion calls in `lib/`. Only 7 have a `Weight` object in hand:

| has a `Weight` (7) | bare kilogram `double` (13) |
|---|---|
| history rows, current-weight hero, bite-page weight line, heatmap readout's two weigh-ins, vs-low's `lowest` | weekly change / 30-day trend / vs-low deltas, the change indicator's `diff`, both chart axes and tooltips, the daily-bites overlay and its semantics, the dialog's prefill and its save |

The deltas and axis values are never `Weight`s — a change isn't a weigh-in — and the dialog converts the *other* direction, from typed input into kilograms. So the `MeasurementSystem` extension has to stay for those 13, and adding a `Weight` one gives two spellings of the same operation.

**Three ways to go, and I'm happy with any:**

1. **Leave it.** One vocabulary, works everywhere.
2. **Add the `Weight` extension as a thin wrapper** over the existing one, used at those 7 sites. ~10 lines, reads better where it fits, costs the second spelling. I can push it in a few minutes if you like it.
3. **If the threading is the real itch**, it's the context-read version that removes it — widgets fetch the system themselves instead of taking it as a parameter. That one I'd still do as a separate change covering weight and height together.

Your call — say which and I'll do it.

---

PR #137, opened 2026-09-13
feat(weight): delete a weight only from the edit dialog

### LINKED ISSUE #136: feat(weight): don't delete weight by swiping
Remove the option to delete a weight by swiping, only through dialog.

igorbasko01:
Swipe-to-delete is gone from the Weight tab's history list. The edit dialog's **Delete** action — which already existed — is now the only way to remove a weigh-in.

## What changed

- `lib/ui/pages/weight_page.dart` — the history row's `Dismissible` wrapper (its red background, delete icon, and `onDismissed` → `deleteWeight` call) is removed; the `ListTile` it wrapped is returned directly and keeps the `ValueKey(item.date)` so the list still identifies rows by day.
- Tapping a row still opens `AddWeightDialog`, and `_showAddWeightDialog`'s existing `result['delete'] == true` branch still calls `manager.deleteWeight`. That path is untouched.

Why this is worth doing: the swipe deleted immediately, with no confirmation and no undo, so a stray horizontal drag on a vertically-scrolling list silently lost an entry. The dialog route is deliberate — open the row, press Delete — and it is the one the app already shipped.

## Meeting the issue

The issue is one line: *"Remove the option to delete a weight by swiping, only through dialog."*

| Asked for | How it is met |
| --- | --- |
| Swipe no longer deletes | The `Dismissible` is gone; a horizontal drag on a row now does nothing. Covered by a test that drags a row 500px and asserts the entry, its row, and `find.byType(Dismissible)` afterwards |
| Deleting still possible through the dialog | The dialog's Delete button and the page's delete branch are unchanged. Covered by a test that taps a row, presses Delete, and asserts the row and the entry are gone |

## Decisions

- **Nothing was added to replace the swipe.** No trailing delete icon on the row, no long-press menu, no confirmation dialog. The issue says deletion happens through the dialog, and that affordance already exists (the row carries an `edit_outlined` icon signalling it is tappable), so adding a second route would go beyond the stated scope.
- **`ValueKey(item.date)` moved onto the `ListTile`** rather than being dropped with the `Dismissible`. It is no longer required — only `Dismissible` demanded it — but keeping it preserves per-day element identity in the `SliverChildBuilderDelegate` as the history range changes, which is free and matches the one-entry-per-day invariant.
- **No confirmation step was introduced.** Removing the accidental-deletion path is the whole point of the issue; gating the deliberate one would be a separate change.

## Verification

Flutter is not installed in this environment and `CLAUDE.md` says not to install it, so `flutter analyze` and `flutter test` were **not** run locally — verification is deferred to the `flutter_ci.yml` checks on this PR. The diff is small and was self-reviewed instead.

Two tests added to `test/ui/pages/weight_page_test.dart`:

- `swiping a history row leaves the entry alone` — drags the row horizontally and asserts no `Dismissible` exists, the row still renders, and `manager.history` still holds the entry.
- `the row dialog is still a way to delete an entry` — taps the row, presses Delete, and asserts the row is gone and `manager.history` is empty.

Both use `InMemoryWeightRepository`, as the existing tests in that file do. No existing test covered swipe-to-delete, so nothing was removed or weakened to make this pass.

Closes #136

https://claude.ai/code/session_019jZrRDqUdHNj2s1WZcHZJb

---
_Generated by [Claude Code](https://claude.ai/code/session_019jZrRDqUdHNj2s1WZcHZJb)_


igorbasko01:
**Verdict: nothing blocks merge.** The change does exactly what the issue asks, is scoped to the one widget that needed it, and the two new tests actually pin the behaviour in both directions. CI (`analyze_and_test`) is green on `3084f05`.

### What I checked against this repo's invariants

None of the usual hazards are in play here, and I confirmed that rather than assuming it:

- **No persistence surface touched.** The diff is two files — `lib/ui/pages/weight_page.dart` and its mirrored test. No Hive model, `typeId`, `@HiveField` index, `@DriftDatabase` table, `schemaVersion`, or `*.g.dart` is involved, so there is no stale-generated-code or migration question.
- **Day granularity intact.** `ValueKey(item.date)` was kept on the `ListTile` rather than dropped with the `Dismissible`. Dates in `history` are already day-normalized and one-entry-per-day, so the keys stay unique and per-day element identity survives a history-range change. Keeping it was the right call even though only `Dismissible` strictly required it.
- **Mutation shape unchanged.** The surviving delete route still goes `manager.deleteWeight` → `_weightRepository.deleteWeight` → `_reloadHistory()`. Nothing mutates the in-memory `_weights` list directly.
- **Backup/versioning untouched.** No codec, no `SerializationService` change, no manual `pubspec.yaml` / `CHANGELOG.md` bump.
- **Title.** `feat(weight): …` is a correct Conventional Commit prefix for a user-visible behaviour change; release-please will cut a minor, which is right for this.

### Tests

Both new tests are genuine rather than decorative:

- `swiping a history row leaves the entry alone` — a 500px horizontal drag well exceeds touch slop, so the `ListTile`'s tap recognizer loses and no dialog opens. That matters: had the drag been read as a tap, the dialog's date button would render the same date string and `find.text(fullDateWithWeekday(today))` would find two widgets, failing `findsOneWidget`. The test genuinely distinguishes "swipe does nothing" from "swipe accidentally taps".
- `the row dialog is still a way to delete an entry` — `find.widgetWithText(TextButton, 'Delete')` is unambiguous: `Cancel` is the only other `TextButton` action and `Save` is a `FilledButton`, which is not a `TextButton` subclass.
- Both use `InMemoryWeightRepository`, matching the rest of the file, and the test lives at `test/ui/pages/weight_page_test.dart`, mirroring `lib/`.

I also grepped the tree: `Dismissible` now appears nowhere in `lib/`, no other caller of `deleteWeight` offered a swipe affordance, and no doc or test elsewhere referenced swipe-to-delete, so nothing was left dangling and no existing test had to be weakened.

### One non-blocking observation (pre-existing)

`lib/ui/pages/weight_page.dart:141-152` — `_showAddWeightDialog` is declared `void … async` and calls `manager.deleteWeight(weight.date)` without awaiting it. If that future ever rejects, the error becomes an unhandled async error with no UI feedback. This is pre-existing and untouched by the diff; I mention it only because this PR makes that branch the *sole* path to deleting a weigh-in, so it is now the one place where a silent failure would leave the user with no alternative. Fine to leave for a separate change.

---

igorbasko01:
please merge from main and resolve conflicts.

igorbasko01:
Merged `main` in (`1cf15a5`) and resolved the conflict. `analyze_and_test` is green on the merge commit and the PR is mergeable again.

The conflict was in `lib/ui/pages/weight_page.dart`, on the history row — #135 had reworked exactly the lines this PR deletes. Both sides kept:

- **From `main`:** the row's weight now renders through `system.formatWeight(item.value)` instead of a hardcoded `'${item.value.toStringAsFixed(1)} kg'`, and the tap handler passes `system` into `_showAddWeightDialog(context, weightManager, system, weight: item)`.
- **From this branch:** the `Dismissible` wrapper is still gone, so the `ListTile` is returned directly and carries the `ValueKey(item.date)`.

Nothing else in the merge conflicted; `test/ui/pages/weight_page_test.dart` merged cleanly, and the two tests this PR adds still compile against the reworked `pumpPage` (its new `system` parameter defaults to metric, so the `'72.5 kg'` assertions hold).

`git diff origin/main -- lib/` is now just the `Dismissible` removal — no part of #135 was reverted in the resolution.

---

PR #138, opened 2026-09-14
feat: show current BMI on a banded scale (underweight / healthy / overweight / obese)

### LINKED ISSUE #130: feat: show current BMI on a banded scale (underweight / healthy / overweight / obese)
## Problem

The Weight tab shows numbers with no reference frame: `78.4 kg` says nothing about whether that's a healthy weight for *this* person. BMI is the cheap, conventional way to give the number a frame — and a bar with the standard bands, with a marker on it, communicates it far faster than a bare figure does.

## Depends on

**#129 — height input.** BMI is weight ÷ height², so there is nothing to compute until a height is stored. This issue should not ship a height field of its own.

## Work

**Computation** — a pure `lib/features/weight/data/bmi.dart`, no widgets, no repository:

- `bmi = kg / (m * m)`, from the most recent weight entry and the stored height in cm. Weight is already kg internally (#120); pre-#120, stored values are treated as kg regardless of `Weight.unit`, consistent with that issue's import rule.
- A `BmiCategory` enum with the WHO adult cut-offs: `underweight` (< 18.5), `healthy` (18.5–25), `overweight` (25–30), `obese` (≥ 30).
- **Half-open intervals** — `[18.5, 25)` — so every value lands in exactly one band and 25.0 is unambiguously overweight. This is worth stating in code, because "18.5–24.9" phrasing leaves 24.95 homeless.

**The scale widget** — a horizontal bar in `lib/ui/widgets/`:

- Four coloured segments, sized in proportion to their BMI ranges, with a marker at the current value and a caption reading the value to one decimal plus the band name (`24.1 · Healthy`).
- Fixed colour constants per band with a `BmiBandStyle.of(category)` lookup, mirroring `PacingZoneStyle` (`lib/ui/widgets/pacing_zone_style.dart`) — including its foreground/background contrast pairing, so band labels stay legible and the semantics read the same in either theme.
- **The rendered domain is clamped** (roughly 15–40); the open-ended obese band would otherwise have no right edge and the healthy band would be a sliver. A BMI outside the domain pins the marker at the end of the bar while the caption still shows the true value.

**Placement and states**

- On the Weight tab, beneath the stat-tile row — note #119 is reworking that row, so coordinate rather than assume its current shape.
- **No height set** → the scale is replaced by a prompt to set a height, linking to Settings. Do not guess a height.
- **No weight entries** → a placeholder, same as the rest of the tab's empty state.

**Accessibility**

- One spoken phrase per the `StatTile` pattern (`Semantics(excludeSemantics: true)`): "BMI 24.1, healthy range".
- The band must never be carried by colour alone — the category is always named in text.

## Notes

- **BMI is an adult screening heuristic, not a diagnosis.** It ignores body composition, so it misreads muscular builds, and it isn't valid for children or during pregnancy. A single quiet line under the scale saying as much; this issue deliberately does **not** implement paediatric percentile curves.
- Which weight feeds it: the latest entry, so the scale tracks the same number the rest of the tab shows. A 7-day average would be steadier but would disagree with the tiles beside it — not worth the confusion.
- Tests: category boundaries at exactly 18.5 / 25 / 30, the clamped domain at both ends, marker position for a mid-band value, and widget tests for the unset-height and no-entries states.
- Related: #129 (height, blocking), #120 (kg as the internal unit), #119 (the stat-tile row this sits under).

igorbasko01:
Puts the latest weigh-in on the four WHO BMI bands on the Weight tab, so the figure has a reference frame instead of standing alone.

## What changed

- **`lib/features/weight/data/bmi.dart`** — pure computation, no widgets and no repository: `bodyMassIndex(kilograms:, heightCm:)`, a `BmiCategory` enum, `bmiCategory()`, and `BmiBand.all` describing each band's interval for a caller that draws them. Cut-offs are named constants (`healthyBmiFrom` 18.5, `overweightBmiFrom` 25, `obeseBmiFrom` 30) and the half-open rule is stated in the code.
- **`lib/ui/widgets/bmi_band_style.dart`** — a fixed `fillColor` / `onFillColor` / `label` per band with `BmiBandStyle.of(category)`, mirroring `PacingZoneStyle` including its contrast pairing (dark text on the amber band, white on the rest).
- **`lib/ui/widgets/bmi_scale.dart`** — the bar: four segments sized in proportion to the ranges they cover, a marker over the current value, and a caption reading the value to one decimal plus the band name (`24.1 · Healthy`), with the adult-screening caveat beneath it.
- **`lib/ui/pages/weight_page.dart`** — the scale sits directly beneath the stat-tile row, with the two empty states.

No Hive model, Drift table, generated file, or dependency wiring is touched — `SettingsManager` is already provided in `lib/main.dart`.

## Acceptance criteria

| From the issue | How it is met |
| --- | --- |
| Pure `bmi.dart`, `kg / m²` from the latest entry and the stored cm | `bodyMassIndex` converts cm → m; the page feeds it `weightManager.latestEntry`. Stored values are read as kg whatever their `WeightUnit` says, per #120's import rule. |
| `BmiCategory` with the WHO cut-offs | `underweight` < 18.5, `healthy` 18.5–25, `overweight` 25–30, `obese` ≥ 30. |
| Half-open intervals, 25.0 unambiguously overweight | `bmiCategory` tests `bmi < bound` upward, so each value lands in exactly one band; asserted at 18.49/18.5, 24.99/25.0, 29.99/30.0. |
| Four proportional segments, marker, caption `24.1 · Healthy` | `BmiScale` gives each segment `flex` from its drawn span in tenths of a point (3.5 / 6.5 / 5 / 10), with an `Align`ed marker and that caption. |
| Fixed band colours via a `BmiBandStyle.of` lookup mirroring `PacingZoneStyle` | Same shape and same contrast pairing; a test asserts a >3:1 WCAG ratio for every band, as `pacing_zone_style_test.dart` does. |
| Clamped domain (~15–40), marker pinned, caption still true | `domainFrom` 15 / `domainTo` 40; `markerFraction` clamps to [0, 1] while the caption prints the real figure (`52.3 · Obese`). |
| Beneath the stat-tile row | Its own sliver right after the tile row — the row reworked in #134 is untouched. |
| No height → a prompt, no guessed height | The scale gives way to "Set your height to see your BMI." plus a **Set height** button. |
| No weight entries → a placeholder | "Log a weigh-in to see your BMI." |
| One spoken phrase, per the `StatTile` pattern | `Semantics(excludeSemantics: true)` reading "BMI 24.1, healthy range". |
| Band never carried by colour alone | Every segment is labelled in text, and the caption names the current band. |
| The adult-screening caveat | One quiet line under the scale; no paediatric percentile curves. |

- **The height prompt opens the Settings height dialog directly** rather than navigating to the Settings tab. The shell is an `IndexedStack` with no cross-tab navigation, and adding that plumbing would be larger than this issue. The button reuses the same `HeightDialog`, in the same measurement system, writing through the same `SettingsManager` — so the height is still answered in one place.
- **The band is chosen from the figure as printed**, i.e. rounded to one decimal first. 81 kg at 180 cm computes to 24.999999999999996 in floating point; without this it would caption "25.0 · Healthy" and contradict itself. This follows `weight_page.dart`'s existing `_roundTo` rule, where a tile's sign and arrow agree with its digits.
- **Segments carry a short label** (`Under` / `Over`) alongside the full name used in the caption and semantics: the underweight segment is a seventh of the bar, and a scaled-down "Underweight" there would be unreadable.
- **The drawn domain is 15–40**, giving spans of 3.5 / 6.5 / 5 / 10 points — the healthy band stays a legible quarter of the bar.
- **Blue for underweight**, with the pacing palette's green / amber / red reused for the rest, so the three bands the two features share read the same.

## Verification

Flutter is not installed in this environment, and per `CLAUDE.md` it was not installed — verification is deferred to the `flutter_ci.yml` checks (`flutter analyze` + `flutter test`) on this PR. The diff was self-reviewed instead; I'll drive the checks green.

Tests added:
- `test/features/weight/bmi_test.dart` — the computation, the band boundaries at exactly 18.5 / 25 / 30, the open ends, and `BmiBand.all` covering every value once.
- `test/ui/widgets/bmi_scale_test.dart` — `markerFraction` across and past the domain, the caption, the print-then-band rule, the marker's position mid-domain and pinned at the edge, segment widths in proportion, all four band names in text, and the spoken phrase.
- `test/ui/widgets/bmi_band_style_test.dart` — distinct labels/fills and a real contrast ratio per band.
- `test/ui/pages/weight_page_test.dart` — the scale from a weigh-in plus a stored height, the unset-height prompt (including setting a height through it and the scale appearing), and the no-entries placeholder. Its `pumpPage` now also provides a `SettingsManager`.

Closes #130

https://claude.ai/code/session_01UyWFBgfqC64CgLtho36R5g

---
_Generated by [Claude Code](https://claude.ai/code/session_01UyWFBgfqC64CgLtho36R5g)_


igorbasko01:
**Verdict: nothing blocks merge.** The change stays clear of every store-level invariant — no Hive model or `typeId`, no Drift table or `schemaVersion`, no generated file, no backup codec (BMI is derived at read time from the latest entry and the stored height, so there is correctly nothing new to serialize), no day-granularity path, and no manual version or `CHANGELOG.md` bump. The height write goes through `SettingsManager` → repository → `notifyListeners`, which is the mutation shape the repo asks for, and `feat:` is the right Conventional Commit prefix for a user-visible addition. Two things worth fixing before or after merge, neither fatal.

### 1. A `WeightUnit.pounds` entry is read as kilograms, and that changes the band, not just a suffix — `lib/ui/pages/weight_page.dart:319`

```dart
return BmiScale(
  bmi: bodyMassIndex(kilograms: current.value, heightCm: heightCm),
);
```

`current.unit` is never consulted. The PR body calls this "#120's import rule", but pounds entries are a live state elsewhere in this codebase, not a dead enum value: `WeightBackupCodec.parseWeightCsv` deliberately preserves `unit=pounds` on import (guarded by `test/features/settings/data/weight_backup_codec_test.dart:78`), `hive_ce_migration_test.dart:43` stores a pounds entry as part of the compatibility contract, `WeeklyWeightChange.level` doubles its thresholds for pounds, and `WeightChangeIndicator` renders an `lbs` suffix.

Concrete case: restore a backup containing `2023-10-27T00:00:00.000,166.0,pounds` (that exact row is in the codec test) with a 175 cm height. `bodyMassIndex` gets 166.0 as kilograms and returns 54.2, and the scale captions **"54.2 · Obese"** for someone with a BMI of 24.6. The rest of the Weight tab already assumes kg (`_CurrentWeight` hard-codes `' kg'`), so the assumption is inherited — but a wrong unit suffix is cosmetic, whereas a wrong BMI band is a health claim the user has no way to see through. Worth either converting on the way in (`lib/core/units.dart` has no mass constant yet, so this is a new `kilogramsPerPound = 0.45359237`) or, if pounds really are considered legacy, falling back to the "can't compute" note rather than silently computing a wrong band.

### 2. The marker is offset by up to half its own icon width, so it can point at a band the caption does not name — `lib/ui/widgets/bmi_scale.dart:63`

```dart
Align(
  alignment: Alignment(markerFraction(bmi) * 2 - 1, 0),
  child: Icon(Icons.arrow_drop_down, size: 28, ...),
),
```

`Align` positions the child *inside* the parent's box, so the 28 px icon's centre sweeps from `14` to `W − 14`, not from `0` to `W`. The drawn centre is `14 + f·(W−28)` where the bar's true position is `f·W` — an error of `14 − 28f` px, zero only at the midpoint and largest at the ends.

Concrete case, on a 360 dp phone (bar width 328 dp after the page's 16 dp padding): the underweight segment ends at 45.9 px. A BMI of 18.4 has `f = 0.136`, true position 44.6 px — just inside blue — but is drawn at **54.8 px**, over the green Healthy segment, while the caption reads "18.4 · Underweight". That is ~0.78 BMI points of drift there, and ~1.07 points at either end of the domain.

The existing test doesn't catch this because it only asserts the midpoint (`27.5`, where the error is exactly 0, at 1.0 px tolerance) and the clamped edge (at 16 px tolerance — which is precisely the 14 px offset being absorbed). A `LayoutBuilder` + `Stack`/`Positioned(left: f * constraints.maxWidth - 14)`, or wrapping the icon in a `FractionalTranslation(translation: Offset(-0.5, 0))` inside a fractional-offset layout, fixes it; then the edge test can tighten to the same 1 px tolerance as the midpoint one.

### Nits

- `_Bmi._editHeight` (`weight_page.dart:325`) is a verbatim copy of `_SettingsPageState._editHeight` (`settings_page.dart:293`) — same dialog, same system, same write. Fine as-is, but a static `HeightDialog.edit(BuildContext)` would keep the two from drifting apart when the height editor grows a third caller.
- The band is picked from the rounded `reading` while the marker uses the raw `bmi`. Deliberate per the PR body, and the residual disagreement is under 0.05 BMI points — mentioning it only so it isn't mistaken for the offset above.

### Verified clean

- `bmi.dart` is genuinely pure — no widget, no repository, no Provider — and sits correctly under `features/weight/data`.
- The half-open banding is right, and `bmi_test.dart` pins 18.49/18.5, 24.99/25.0, 29.99/30.0 plus both open ends.
- No divide-by-zero or NaN path into `markerFraction`: `HeightDialog` clamps to 50–272 cm and `ProfileBackupCodec.parseProfileCsv` drops any restored height that isn't `> 0`, so `heightCm` is never 0 or negative.
- `context.read` is taken before the `await` in `_editHeight` and no `BuildContext` is touched after it.
- `_segment`'s flex arithmetic is correct: 35 / 65 / 50 / 100 tenths over the 15–40 domain.
- Tests mirror `lib/` and use `InMemoryWeightRepository` / `InMemorySettingsRepository` throughout.
- At review time `analyze_and_test` was still in progress on this head — per `CLAUDE.md` that's the verification path for this change, so the findings above are from reading, not from a run.

---

igorbasko01:
Thanks — acting on one of the two findings, leaving the other for you to weigh.

**Marker geometry — fixed in e8c5c4f.** Confirmed exactly as described: `Align` swept the glyph's *box*, so the drawn centre was `14 + f·(W − 28)` instead of `f·W`. On the 800 px test bar a BMI of 18.4 drew at 119 px while the underweight segment ends at 112 — the marker sat on the green band under an "Underweight" caption. The marker's left edge is now computed from the bar's width (`_markerLeft`), pinned by the same arithmetic at the ends. Two tests added: a non-midpoint value (20.0 lands a fifth along the bar, where the old and new formulas disagree) and a containment check that 18.4 stays inside the segment its caption names.

**Pounds entries — left as is, deliberately.** #130 settles this in its own words: *"Weight is already kg internally (#120); pre-#120, stored values are treated as kg regardless of `Weight.unit`, consistent with that issue's import rule."* Converting here would also make the scale disagree with the `78.4 kg` hero figure and the history rows directly above it, which hard-code the kg suffix — BMI would become the only place in the tab reading `unit`, showing one number as a weight and a different body as a band. #120 (open in #135) is where that gets fixed for the whole tab, and this scale picks the correction up for free once it lands. Worth weighing there rather than here; flagging it so the decision is visible rather than inherited silently.

**Nits:** `_editHeight` is indeed duplicated with `settings_page.dart` — extracting it touches a file outside this issue's scope, so I've left it. The marker reading raw `bmi` while the band reads the rounded value is intentional (< 0.05 BMI points, and the rounding exists so the caption can't contradict itself).

---

igorbasko01:
merge from main and resolve conflicts

igorbasko01:
Merged `main` in — `771030b`. CI green on the merge head (run 262), `mergeable_state: clean`.

Two conflicts, both in the Weight tab:

- **`lib/ui/pages/weight_page.dart`** — imports only. The BMI sliver and `_Bmi` slot in beneath the stat-tile row #135 reshuffled; the diff against `main` is now purely additive.
- **`test/ui/pages/weight_page_test.dart`** — `pumpPage` took a `settingsManager` override on this branch and a `system` parameter on `main`, so it now takes both, with the default repository carrying `heightCm: 175` *and* the requested system. Kept both sides' tests: #137's swipe/delete cases and the BMI group.

One semantic consequence worth flagging: **#135 landing settles the pounds question from the review above.** Weights are kilograms everywhere now, so `bodyMassIndex(kilograms: current.value, …)` is correct by construction rather than by the pre-#120 convention, and `Weight.unit` is legacy provenance no computation reads. I dropped the comment on `bodyMassIndex` that described the old rule; nothing else changed. BMI itself is unitless, so the new display preference doesn't reach the scale — imperial users get the same number with the same band.

---

igorbasko01:
The **Build Test APK** run you dispatched failed on its first attempt, but not on anything in this branch — I re-ran it and [attempt 2 is green](https://github.com/igorbasko01/food-locker/actions/runs/35617820993). The test APK is published to the rolling `test-builds` release as usual.

The failure was in `package:sqlite3`'s native-asset hook, downloading a prebuilt `.so` whose hash didn't match what the package expects:

```
Bad state: Hash of downloaded file libsqlite3.arm.android.so is bef140a1…,
expected 740bbde8…
  Building assets for package:sqlite3 failed.
```

`pubspec.lock` on this branch is byte-identical to `main`'s, this PR touches no Drift, sqlite3 or Android code, and the same download succeeded on the second attempt from the same commit — so it was a corrupted fetch, not a version or upstream-artifact problem. (`CLAUDE.md` already notes this download is unreliable in sandboxes; worth knowing it can bite the Actions runner too.) That was the one re-run, so if it recurs on a later build it is worth treating as real rather than re-running again.

Flutter CI on this head stays green; `mergeable_state` is clean.

---

